PR #21, opened 2021-06-23
Add JacksonRegistry

snazy:
The new `JacksonRegistry` is an alternative to the `ProtoTypeRegistry`. It allows applications
to pass Java/Jackson objects as arguments to a CEL expression. This change does not support
_returning_ Java/Jackson objects _from_ a CEL expression.
    
The primary use case of this change is to enable CEL scripts to evaluate an expression against
some Java/Jackson object tree and return a CEL built-in type (primitive types like string or
boolean) - TL;DR security policy checks or filter-predicates.

### DIFF
--- a/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
+++ b/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
@@ -36,7 +36,7 @@ import org.projectnessie.cel.common.types.ref.Val;
  *
  * <p>The implementation does not support the construction of Jackson objects in CEL expressions and
  * therefore returning Jackson objects from CEL expressions is not possible/implemented and results
- * in {@link UnsupportedOperationException}s.</p>
+ * in {@link UnsupportedOperationException}s.
  */
 public final class JacksonRegistry implements TypeRegistry {
   final ObjectMapper objectMapper;
--- a/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
+++ b/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.EnumSerializer;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 import java.util.HashMap;
 import java.util.Map;
@@ -44,6 +45,9 @@ public final class JacksonRegistry implements TypeRegistry {
   private final TypeFactory typeFactory;
   private final Map<Class<?>, JacksonTypeDescription> knownTypes = new HashMap<>();
   private final Map<String, JacksonTypeDescription> knownTypesByName = new HashMap<>();
+
+  private final Map<Class<?>, JacksonEnumDescription> enumMap = new HashMap<>();
+  private final Map<String, JacksonEnumValue> enumValues = new HashMap<>();
 
   private JacksonRegistry() {
     this.objectMapper = new ObjectMapper();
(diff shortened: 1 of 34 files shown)


nastra (approved):

jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
```diff
@@ -0,0 +1,138 @@
...
+import org.projectnessie.cel.common.types.ref.Type;
+import org.projectnessie.cel.common.types.ref.TypeAdapterSupport;
+import org.projectnessie.cel.common.types.ref.TypeRegistry;
+import org.projectnessie.cel.common.types.ref.Val;
+
+/**
+ * CEL-Java {@link TypeRegistry} to use Jackson objects as input values for CEL scripts.
+ *
+ * <p>The implementation does not support the construction of Jackson objects in CEL expressions and
+ * therefore not to return Jackson objects from CEL expressions.
```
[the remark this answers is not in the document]
snazy: nice catch!
fixed

nastra (commented):
fyi that having an Enum doesn't work 
```
--- a/jackson/src/test/java/org/projectnessie/cel/types/jackson/types/CommitMeta.java
+++ b/jackson/src/test/java/org/projectnessie/cel/types/jackson/types/CommitMeta.java
@@ -36,9 +36,16 @@ import org.immutables.value.Value;
 @JsonDeserialize(as = ImmutableCommitMeta.class)
 public abstract class CommitMeta {
 
+  public enum Type {
+    ICEBERG, DELTA;
+  }
+
+  @Nullable
+  public abstract Type getType();
```
It fails at:
```
Unsupported Java Type '[simple type, class org.projectnessie.cel.types.jackson.types.CommitMeta$Type]'
java.lang.UnsupportedOperationException: Unsupported Java Type '[simple type, class org.projectnessie.cel.types.jackson.types.CommitMeta$Type]'
	at org.projectnessie.cel.types.jackson.JacksonTypeDescription.findTypeForJacksonType(JacksonTypeDescription.java:114)
	at org.projectnessie.cel.types.jackson.JacksonTypeDescription.<init>(JacksonTypeDescription.java:66)
	at org.projectnessie.cel.types.jackson.JacksonRegistry.computeTypeDescription(JacksonRegistry.java:132)
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1138)
	at org.projectnessie.cel.types.jackson.JacksonRegistry.typeDescription(JacksonRegistry.java:125)
	at org.projectnessie.cel.types.jackson.JacksonRegistry.register(JacksonRegistry.java:66)
	at org.projectnessie.cel.EnvOption.lambda$types$8(EnvOption.java:239)
	at org.projectnessie.cel.Env.configure(Env.java:448)
	at org.projectnessie.cel.Env.newCustomEnv(Env.java:136)
	at org.projectnessie.cel.tools.ScriptHost.getOrCreateScript(ScriptHost.java:65)
	at org.projectnessie.cel.types.jackson.JacksonRegistryTest.scriptHost(JacksonRegistryTest.java:179)
```

rymurr:
Out of curiosity what is Type for in CommitMeta? How can the commit log
have a type? Eg what happens if the commit has multiple types in it?

Is this localised to only cel code? If yes we shouldn't be polluting cel w/
Nesie types, I think a more generic and less Nessie class would be more
suitable for testing.

nastra:
Having an Enum in `CommitMeta` was just an example to show where something is failing. We have Enum types in https://github.com/projectnessie/nessie/blob/52a6141fb3d3b538ecaf1c39006e6160832ec166/model/src/main/java/org/projectnessie/model/EntriesResponse.java#L43 and it's failing the same way

jackson/src/test/java/org/projectnessie/cel/types/jackson/types/Branch.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+/** Api representation of an Nessie Tag/Branch. This object is akin to a Ref in Git terminology. */
+@Value.Immutable(prehash = true)
+@JsonSerialize(as = ImmutableBranch.class)
+@JsonDeserialize(as = ImmutableBranch.class)
+@JsonTypeName("BRANCH")
+public interface Branch extends Reference {
```
rymurr, line 28: Can we not use Nessie objects and make these more widget-y. (same w/ below)

rymurr (changes requested):

snazy:
All the review comments should be addressed now. Enum's should work fine now.

rymurr (approved):

jackson/src/test/java/org/projectnessie/cel/types/jackson/JacksonScriptHostTest.java
```diff
@@ -0,0 +1,101 @@
...
+                    .build())
+            .build();
+
+    assertThat(script.execute(Boolean.class, singletonMap("param", val))).isTrue();
+
+    // same as above, but use the 'container'
+
+    script =
+        scriptHost
+            .buildScript("param.entries[0].type == ClassWithEnum.ClassEnum.VAL_2")
```
nastra, line 78: not sure we really want to make users have to know internal classes in order to apply fliters. Personally I would rather like to see `entry.type == 'ICEBERG_TABLE'` working than having to do `entry.type == org.projectnessie.model.Contents.Type.ICEBERG_TABLE`.

@rymurr thoughts?

rymurr: I wholeheartedly agree!

PR #54, opened 2021-10-29
Relocate antlr-runtime

### LINKED ISSUE #52: Shade antlr-runtime + bump antlr to recent version

snazy:
Fixes #52

### DIFF
--- a/generated-antlr/build.gradle.kts
+++ b/generated-antlr/build.gradle.kts
@@ -47,7 +47,7 @@ tasks.named<ShadowJar>("shadowJar") {
     dependencies {
         include(dependency("org.antlr:antlr4-runtime"))
     }
-    relocate("org.antlr.v4.runtime", "org.projectnessie.cel.parser.gen.antlr.v4.runtime")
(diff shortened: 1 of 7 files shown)


generated-antlr/build.gradle.kts
```diff
@@ -32,3 +35,26 @@ dependencies {
...
+}
+
+tasks.named<ShadowJar>("shadowJar") {
+    // The antlr-plugin should ideally do this
+    dependsOn(tasks.named("generateGrammarSource"))
+
+    dependencies {
+        include(dependency("org.antlr:antlr4-runtime"))
+    }
+    relocate("org.antlr.v4.runtime", "org.projectnessie.cel.parser.gen.antlr.v4.runtime")
```
omarsmak, line 50: Does it make sense if we keep the shaded package name as it is and add it `shaded`? e.g: `org.projectnessie.cel.shaded.org.antlr.v4.runtime`. In order to distinguish that is shaded from `org.antlr.v4.runtime`

snazy: Yup - updated the package name

omarsmak (approved):
LGTM

Issue #103, opened 2022-03-07
Pull in #316 from cel-go: strings extension library


alancao-uber:
Pull in [google/cel-go#316](https://github.com/google/cel-go/pull/316)

PR #104, opened 2022-03-07
port strings extension library from cel-go

alancao-uber:
### issue #103: port strings extension library from [cel-go](https://github.com/google/cel-go/blob/master/ext/strings.go) ### 

- all strings extension functions are covered
- because of limit on [Overload](https://github.com/projectnessie/cel-java/blob/main/core/src/main/java/org/projectnessie/cel/interpreter/functions/Overload.java)'s constructor, multiple overrides for same function are not supported. That will be covered in separated issue/PR
-- e.g. `indexOf` function only accepts one parameter
-- similar to `lastIndexOf`, `replace`, `split`, `substring`
- Gaurds.java is similar to [guards.go](https://github.com/google/cel-go/blob/master/ext/guards.go)
- `TriFunction` and `QuadFunction` interfaces are introduced as JDK doesn't provide [BIFunction]([BiFunction (Java Platform SE 8 ) - Oracle Help Centerhttps://docs.oracle.com › api › java › util › function › B...](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html)) and [UnaryOperator](https://docs.oracle.com/javase/8/docs/api/java/util/function/UnaryOperator.html)

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
+++ b/core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
@@ -1,0 +1,355 @@
+/*
+ * Copyright (C) 2022 The Authors of CEL-Java
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.cel.extension;
+
+import java.util.*;
+import java.util.regex.Pattern;
+import org.projectnessie.cel.EnvOption;
+import org.projectnessie.cel.Library;
+import org.projectnessie.cel.ProgramOption;
+import org.projectnessie.cel.checker.Decls;
+import org.projectnessie.cel.interpreter.functions.Overload;
+
+/**
+ * StringsLib provides a {@link org.projectnessie.cel.EnvOption} to configure extended functions for
+ * string manipulation. As a general note, all indices are zero-based. The implementation is ported
+ * from <a href=https://github.com/google/cel-go/blob/master/ext/strings.go>cel-go</a>.
+ *
+ * <p>Note: Currently the overloading isn't supported.
+ *
+ * <h3>CharAt</h3>
+ *
+ * <p>Returns the character at the given position. If the position is negative, or greater than the
+ * length of the string, the function will produce an error:
+ *
+ * <pre>    {@code <string>.charAt(<int>) -> <string>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'hello'.charAt(4) // return 'o'}</pre>
+ *
+ * <pre>    {@code 'hello'.charAt(5) // return ''}</pre>
+ *
+ * <pre>    {@code 'hello'.charAt(-1) // error}</pre>
+ *
+ * <h3>IndexOf</h3>
+ *
+ * <p>Returns the integer index of the first occurrence of the search string. If the search string
+ * is not found the function returns -1.
+ *
+ * <pre>    {@code <string>.indexOf(<string>) -> <int>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'hello mellow'.indexOf('') // returns 0}</pre>
+ *
+ * <pre>    {@code 'hello mellow'.indexOf('ello') // returns 1}</pre>
+ *
+ * <pre>    {@code 'hello mellow'.indexOf('jello') // returns -1}</pre>
+ *
+ * <h3>LastIndexOf</h3>
+ *
+ * <p>Returns the integer index at the start of the last occurrence of the search string. If the
+ * search string is not found the function returns -1.
+ *
+ * <pre>    {@code <string>.lastIndexOf(<string>) -> <int>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'hello mellow'.lastIndexOf('') // returns 12}</pre>
+ *
+ * <pre>    {@code 'hello mellow'.lastIndexOf('ello') // returns 7}</pre>
+ *
+ * <pre>    {@code 'hello mellow'.lastIndexOf('jello') // returns -1}</pre>
+ *
+ * <h3>LowerAscii</h4>
+ *
+ * <p>Returns a new string where all ASCII characters are lower-cased.
+ *
+ * <p>This function does not perform Unicode case-mapping for characters outside the ASCII range.
+ *
+ * <pre>    {@code <string>.lowerAscii() -> <string>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'TacoCat'.lowerAscii() // returns 'tacocat'}</pre>
+ *
+ * <pre>    {@code 'TacoCÆt Xii'.lowerAscii() // returns 'tacocÆt xii'}</pre>
+ *
+ * <h3>Replace</h3>
+ *
+ * <p>Returns a new string based on the target, which replaces the occurrences of a search string
+ * with a replacement string if present.
+ *
+ * <pre>    {@code <string>.replace(<string>, <string>) -> <string>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'hello hello'.replace('he', 'we') // returns 'wello wello'}</pre>
+ *
+ * <h3>Split</h3>
+ *
+ * <p>Returns a list of strings split from the input by the given separator.
+ *
+ * <pre>    {@code <string>.split(<string>) -> <list<string>>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code hello hello hello'.split(' ')     // returns ['hello', 'hello', 'hello']}</pre>
+ *
+ * <h3>Substring</h3>
+ *
+ * <p>Returns the substring given a numeric range corresponding to character positions.
+ *
+ * <p>Character offsets are 0-based with an inclusive start range.
+ *
+ * <pre>    {@code <string>.substring(<int>) -> <string>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'tacocat'.substring(4) // returns 'cat'}</pre>
+ *
+ * <pre>    {@code 'tacocat'.substring(-1) // error} </pre>
+ *
+ * <h3>Trim</h3>
+ *
+ * <p>Returns a new string which removes the leading and trailing whitespace in the target string.
+ * The trim function uses the Unicode definition of whitespace which does not include the zero-width
+ * spaces. See: <a href="https://en.wikipedia.org/wiki/Whitespace_character#Unicode">Unicode</a>
+ *
+ * <pre>    {@code <string>.trim() -> <string>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code ' \ttrim\n '.trim() // returns 'trim'}</pre>
+ *
+ * <h3>UpperAscii</h3>
+ *
+ * <p>Returns a new string where all ASCII characters are upper-cased.
+ *
+ * <p>This function does not perform Unicode case-mapping for characters outside the ASCII range.
+ *
+ * <pre>    {@code <string>.upperAscii() -> <string>}</pre>
+ *
+ * <h4>Examples:</h4>
+ *
+ * <pre>    {@code 'TacoCat'.upperAscii() // returns 'TACOCAT'}</pre>
+ *
+ * <pre>    {@code 'TacoCÆt Xii'.upperAscii() // returns 'TACOCÆT XII'}</pre>
+ */
+public class StringsLib implements Library {
+
+  private static final String CHAR_AT = "charAt";
+  private static final String INDEX_OF = "indexOf";
+  private static final String LAST_INDEX_OF = "lastIndexOf";
+  private static final String LOWER_ASCII = "lowerAscii";
+  private static final String REPLACE = "replace";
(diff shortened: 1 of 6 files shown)


snazy:
Thank you, this looks really nice!
I'll take a look.

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
```diff
@@ -0,0 +1,298 @@
...
+import org.projectnessie.cel.interpreter.functions.Overload;
+
+/**
+ * StringsLib provides a {@link org.projectnessie.cel.EnvOption} to configure extended functions for
+ * string manipulation. As a general note, all indices are zero-based. The implementation is ported
+ * from <a href=https://github.com/google/cel-go/blob/master/ext/strings.go>cel-go</a>.
+ *
+ * <p>Note: Currently the overloading isn't supported.
+ *
+ * <p>CharAt
```
snazy, line 35: For better readability:
* use e.g. `<h2>` or so for the function names
* surround the code examples with javadoc's `{@code ...}`
* Note that using `<` and `>` for content's not going to work - that'd be interpreted as an HTML tag and eventually not shown at all

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
```diff
@@ -0,0 +1,298 @@
...
+    return list;
+  }
+
+  @Override
+  public List<ProgramOption> getProgramOptions() {
+    List<ProgramOption> list = new ArrayList<>();
+    ProgramOption functions =
+        ProgramOption.functions(
+            Overload.binary(CHAR_AT, Gaurds.callInStrIntOutStr(StringsLib::charAt)),
+            Overload.binary(INDEX_OF, Gaurds.callInStrStrOutInt(StringsLib::indexOf)),
```
snazy, line 212: Nit: these "simple" ones could be replaced with `String::indexOf`

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Test
+  public void testLowerAscii() {
+    testExpression("'TacoCat'.lowerAscii() == 'tacocat'");
+    testExpression("'TacoCÆt Xii'.lowerAscii() == 'tacocÆt xii'");
+  }
+
+  @Test
+  public void testReplace() {
+    testExpression("'hello hello'.replace('he', 'we') == 'wello wello'");
```
snazy, line 73: Can you add the tests from https://github.com/google/cel-go/pull/316/files#diff-01b0150ddfed956af4270fcad9e73cc6f8d8ea0ad279353349682800208cc5fdR54 here?

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @Test
+  public void testSplit() {
+    testExpression("'hello hello hello'.split(' ') == ['hello', 'hello', 'hello']");
```
snazy, line 78: Would be nice to have the tests from https://github.com/google/cel-go/pull/316/files#diff-01b0150ddfed956af4270fcad9e73cc6f8d8ea0ad279353349682800208cc5fdR59 here

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Test
+  public void testUpperAscii() {
+    testExpression("'TacoCat'.upperAscii() == 'TACOCAT'");
+    testExpression("'TacoCÆt Xii'.upperAscii() == 'TACOCÆT XII'");
+  }
+
+  @Test
+  public void testTrim() {
+    testExpression("'  \\ttrim\\n    '.trim() == 'trim'");
```
snazy, line 94: Similar - tests from https://github.com/google/cel-go/pull/316/files#diff-01b0150ddfed956af4270fcad9e73cc6f8d8ea0ad279353349682800208cc5fdR71

core/src/main/java/org/projectnessie/cel/extension/Gaurds.java
```diff
@@ -0,0 +1,202 @@
...
+import org.projectnessie.cel.interpreter.functions.UnaryOp;
+
+/** function invocation guards for common call signatures within extension functions. */
+public class Gaurds {
+
+  private Gaurds() {}
+
+  public static BinaryOp callInStrIntOutStr(BiFunction<String, Integer, String> func) {
+    return (lhs, rhs) -> {
+      if (!(lhs instanceof StringT)) {
```
snazy, line 34: You can safely use `lhs.type() == StringT.StringType` - those are singletons

core/src/main/java/org/projectnessie/cel/extension/Gaurds.java
```diff
@@ -0,0 +1,202 @@
...
+        throw new RuntimeException();
```
snazy, line 35: Throwing an exception isn't great here. Better return the appropriate error-value from one of the helpers in `org.projectnessie.cel.common.types.Err`

XN137:
if i understand this correctly then we might also want to adjust https://github.com/projectnessie/cel-java#not-yet-implemented to no longer state that the strings extension is not yet implemented

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
alancao-uber: It will simpler to use static function as `BinaryOp` or `FunctionOp` etc. And the code style will be consistent to other functions. It could be acceptable to wrap `String.indexOf` to a static function.

core/src/main/java/org/projectnessie/cel/extension/Gaurds.java
alancao-uber: Thanks. That has been addressed

core/src/main/java/org/projectnessie/cel/extension/Gaurds.java
alancao-uber: Thanks. That has been addressed

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
alancao-uber: Done

alancao-uber:
> if i understand this correctly then we might also want to adjust https://github.com/projectnessie/cel-java#not-yet-implemented to no longer state that the strings extension is not yet implemented

This PR doesn't cover all strings extension as cel-go. Like multiple overloadings. Only the first is supported. 
```
"'tacocat'.indexOf('a')
"'tacocat'.indexOf('a', 3)
```

Such case will be handled in separate PR. We can announce supporting strings extension function at that moment.

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
snazy: I'm okay with this.

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
```diff
@@ -0,0 +1,340 @@
...
+  private static final String SPLIT = "split";
+  private static final String SUBSTR = "substring";
+  private static final String TRIM_SPACE = "trim";
+  private static final String UPPER_ASCII = "upperAscii";
+
+  // whitespace characters definition from
+  // https://en.wikipedia.org/wiki/Whitespace_character#Unicode
+  private static final Set<Character> UNICODE_WHITE_SPACES =
+      new HashSet<>(
+          (Collection<Character>)
```
snazy, line 169: Nit: unnecessary cast

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
```diff
@@ -0,0 +1,340 @@
...
+  private static final String LOWER_ASCII = "lowerAscii";
+  private static final String REPLACE = "replace";
+  private static final String SPLIT = "split";
+  private static final String SUBSTR = "substring";
+  private static final String TRIM_SPACE = "trim";
+  private static final String UPPER_ASCII = "upperAscii";
+
+  // whitespace characters definition from
+  // https://en.wikipedia.org/wiki/Whitespace_character#Unicode
+  private static final Set<Character> UNICODE_WHITE_SPACES =
```
snazy, line 167: Can you add a comment why `Character.isWhitespace()` doesn't work for CEL here?

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,349 @@
...
+    testExpression(
+        new TestData(
+            "\"42\".replace(\"2\", \"1\", \"1\") == \"41\"", "no matching overload", true));
+    testExpression(
+        new TestData(
+            "\"42\".replace(\"2\", \"1\", 1, false) == \"41\"", "no matching overload", true));
+  }
+
+  @Test
+  public void testSplit() {
```
snazy, line 145: Looks like there are lot of duplicate tests from other test functions in this one.

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,349 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.cel.Env;
+import org.projectnessie.cel.Library;
+import org.projectnessie.cel.Program;
+import org.projectnessie.cel.common.types.pb.ProtoTypeRegistry;
+
+public class StringsTest {
```
snazy, line 28: Thanks for adding all the tests!

Can you change the repeated calls to testExpression() to parameterized tests like this:

```
  @ParameterizedTest
  @MethodSource("testCharAt")
  public void testCharAt(TestData testData) {
    testExpression(testData);
  }

  static Stream<Arguments> testCharAt() {
    return Stream.of(
      new TestData(...),
      new TestData(...)
    );
  }
```

This will run every test case, also those after a failed one.

As an alternative, you can also wrap all `testExpression` invocations in a `assertAll` like this:
```
    assertAll(
            () -> testExpression(...),
            () -> testExpression(...),
            );
```

I'd be fine with either.

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    public String getErr() {
+      return err;
+    }
+
+    public boolean isParseOnly() {
+      return parseOnly;
+    }
+  }
```
snazy, line 348: Nit: add
```
    @Override
    public String toString() {
      return expression;
    }
```

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,349 @@
...
+      return;
+    } else {
+      assertThat(checked.hasIssues()).isFalse();
+    }
+
+    Program program = env.program(astIssue.getAst());
+    Program.EvalResult result = program.eval(new HashMap<>());
+
+    if (testData.getErr() != null) {
+      assertThat(result.getVal().toString()).isEqualTo(testData.getErr());
```
snazy, line 312:
```suggestion
      assertThat(result.getVal()).extracting(Val::toString).isEqualTo(testData.getErr());
```

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
```diff
@@ -0,0 +1,349 @@
...
+    } else {
+      assertThat(result.getVal().booleanValue()).isTrue();
```
snazy, line 314:
```suggestion
      assertThat(result.getVal()).isEqualTo(BoolT.True);
```

snazy (commented):
Very close! :)
Thanks for the work so far.
A few nits and we're good to merge it (and can publish a new release).

snazy:
Hm. Looks like there's different behavior for Java 17 compared to Java 11 (tests pass with Java 11 locally).

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
```diff
@@ -0,0 +1,340 @@
...
+      start++;
+    }
+    while (end > start) {
+      if (!UNICODE_WHITE_SPACES.contains(chars[end])) {
+        break;
+      }
+      end--;
+    }
+
+    return str.substring(start, end + 1);
```
snazy, line 326: Add `try-catch(IndexOutOfBoundsException)` and throw with a non-Java-version dependent message

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
```diff
@@ -0,0 +1,340 @@
...
+            Overload.unary(UPPER_ASCII, Guards.callInStrOutStr(StringsLib::upperASCII)));
+    list.add(functions);
+    return list;
+  }
+
+  static String charAt(String str, int index) {
+    if (str.length() == index) {
+      return "";
+    }
+    return String.valueOf(str.charAt(index));
```
snazy, line 274: Add `try-catch(IndexOutOfBoundsException)` and throw with a non-Java-version dependent message

snazy:
Mind updating the [README/"not yet implemented"](https://github.com/projectnessie/cel-java#not-yet-implemented) as well?

core/src/test/java/org/projectnessie/cel/extension/StringsTest.java
alancao-uber: Thanks, this is very useful to simplify the test code

core/src/main/java/org/projectnessie/cel/extension/StringsLib.java
alancao-uber: I wonder if we should wrap the Java exception with cel specific message (actually to be same as Go)

For the test cases, another option will be checking the returned value is `Err` type and cause exception is in same class. How do you think?

```
+      assertThat(((Err) result.getVal()).getCause().getClass().getCanonicalName()).isEqualTo(testData.getExceptionClass());
-      assertThat(result.getVal()).extracting(Val::toString).isEqualTo(testData.getErr());
```

snazy: > I wonder if we should wrap the Java exception with cel specific message (actually to be same as Go)

+1

snazy (approved):
LGTM!
Thank you for the contribution!

PR #105, opened 2022-03-08
Turn off codecov github annotations

XN137:
See also https://docs.codecov.com/docs/github-checks

It can be argued that having theses inline annotations is too noisy for reviewers when you already have an automated codecov comment outlining the coverage numbers for the codebase as well as on the diff of the PR.
If you need more line-by-line details you can simply follow the link from the comment to the codecov report.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,20 @@
+#
+# Copyright (C) 2022 The Authors of CEL-Java
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


laurentgo:
Your description probably deserves better than X asked me to disable it because it was annoying to him?

What part is "annoying"? Is the message inaccurate, or do we simply not want to see when code is being added without being tested?

XN137:
I'll defer this discussion to @snazy since he just asked me to put up the PR as a favor.
I am fine either way.

Side note:
main repo already has them turned off (also not my choice)
https://github.com/projectnessie/nessie/blame/8ba7079c54325c90e39ebf4e234c014711ca0e8e/codecov.yml#L30-L31

snazy:
@XN137 please update the PR description to be objective and describe what the PR's doing.

XN137:
ok i tried updating the description

snazy (approved):

PR #123, opened 2022-03-23
Add Java 18 to CI

snazy:


XN137 (approved):
lgtm

just wondering:
is the strategy to really support non LTS java versions or do we just want to keep the codebase in shape ahead of the next java LTS release?

snazy:
It's just to stay ahead and proactively prevent "surprises".

PR #154, opened 2022-04-11
Adopt release-workflow to new dependabot-automerge commit status

snazy:
#153 updated the wrong file :facepalm:


XN137 (approved):
lgtm

should we undo the changes in the other file for consistency with the other repos?

PR #156, opened 2022-04-12
Use the `NESSIE_BUILDER` secret to auto-merge dependabot PRs

snazy:
... because the default `GITHUB_TOKEN` does not trigger other workflows,
like CI on the main branch.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -85,7 +85,7 @@ jobs:
       - name: "Merge pull request"
         uses: "actions/github-script@v6"
         with:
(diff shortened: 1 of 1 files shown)


XN137 (approved):

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: "Merge pull request"
         uses: "actions/github-script@v6"
         with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          github-token: "${{ secrets.NESSIE_BUILDER }}"
```
XN137, line 88: maybe add comment why we need a different secret here, compared to the other places in this file

PR #201, opened 2022-06-10
Fix Java 8 compatibiltiy

snazy:
Although the CEL source code was always compatible with Java 8, the
compiled classes were not, because they were never build with
`--release 8`.

This change adds the changes to Java compilation.

CI didn't catch this issue bug, because the while build was run
with Java 8 and that caused the code to be _built_ again against
Java 8.

So this change also updates the CI workflow to run the tests against
the classes built by Java 11.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,17 +32,46 @@ jobs:
         with:
           submodules: 'true'
 
+      # Special handling to test against Java 8:
+      # antlr version 4.10 requires Java 11 to run the antlr tool, but tests still need to run
+      # against Java 8 - that's being done here:
+      - name: Set up Java 8
+        if: ${{ matrix.java-version == '11' }}
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: 8
+      - name: Configure env with Java 8 home
+        if: ${{ matrix.java-version == '11' }}
+        run:
+          echo "JAVA_8_HOME=$JAVA_HOME" >> $GITHUB_ENV
+      - name: Enable Java 8 tests
+        if: ${{ matrix.java-version == '11' }}
+        run:
+          echo "ADDITIONAL_GRADLE_OPTS=-PalsoTestAgainstJava8" >> $GITHUB_ENV
+      - name: Configure Java 8 home for Gradle
+        if: ${{ matrix.java-version == '11' }}
+        run: |
+          mkdir -p $HOME/.gradle
+          echo "org.gradle.java.installations.fromEnv=JAVA_8_HOME" >> $HOME/.gradle/gradle.properties
+
       - name: Set up JDK ${{ matrix.java-version }}
         uses: actions/setup-java@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -32,17 +32,46 @@ jobs:
...
+          # Spotless must run in a different invocation, because
+          # it has some weird Gradle configuration/variant issue
+          arguments: spotlessCheck
+
       - name: Build with Gradle
-        # Java 8 and 16+ don't run spotless, because google-java-format requires Java 11+ (but doesn't run w/ Java 16)
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: --rerun-tasks --info assemble check publishToMavenLocal -x jmh
+          arguments: --rerun-tasks assemble ${{ env.ADDITIONAL_GRADLE_OPTS }} check publishToMavenLocal -x jmh -x spotlessCheck
```
XN137, line 74: question: do the opts need to be inbetween task names? or could they come after `--rerun-tasks` instead?

build.gradle.kts
```diff
@@ -88,6 +88,28 @@ allprojects {
       useJUnitPlatform {}
       maxParallelForks = Runtime.getRuntime().availableProcessors()
     }
+
+    if (project.hasProperty("alsoTestAgainstJava8")) {
+      afterEvaluate {
+        val javaToolchains = extensions.findByType(JavaToolchainService::class.java)
+        if (javaToolchains != null) {
+          val testWithJava8 =
+            tasks.register<Test>("testWithJava8") {
```
XN137, line 97: doesnt this mean that gradle execution in CI should mention a `testWithJava8` task being executed?
i am not seeing it in the CI run on this PR :thinking:

XN137 (approved):
+1 assuming `testWithJava8` task gets executed in CI (even though i dont see it in logs atm)

.github/workflows/main.yml
```diff
@@ -32,17 +32,46 @@ jobs:
...
+          java-version: 8
+      - name: Configure env with Java 8 home
+        if: ${{ matrix.java-version == '11' }}
+        run:
+          echo "JAVA_8_HOME=$JAVA_HOME" >> $GITHUB_ENV
+      - name: Enable Java 8 tests
+        if: ${{ matrix.java-version == '11' }}
+        run:
+          echo "ADDITIONAL_GRADLE_OPTS=-PalsoTestAgainstJava8" >> $GITHUB_ENV
+      - name: Configure Java 8 home for Gradle
```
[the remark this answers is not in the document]
snazy: IDK why, but that doesn't work. The 2nd echo is appended to the first env's value :facepalm:

.github/workflows/main.yml
snazy: The order doesn't matter in this case.

build.gradle.kts
snazy: Yea - already fixed that

PR #202, opened 2022-06-10
CI: Run conformance tests only once

snazy:
It's just not necessary to run the conformance tests against multiple
different Java versions, because those tests do not hit any code path
in CEL-Java that any way dependends on the Java version.

Running the conformance tests only for one Matrix job also improves
CI runtime, because Bazel's cache depends on the Java version.


XN137 (approved):
yaml changes lgtm but i cant judge whether conformance tests could show different behavior depending on the java version or not, so have to trust you on this one ;)

PR #203, opened 2022-06-10
Fix spotless to include *.kts

snazy:


build.gradle.kts
```diff
@@ -270,6 +270,31 @@ allprojects {
     }
   }
 
+  plugins.withType<SpotlessPlugin>().configureEach {
+    configure<SpotlessExtension> {
+      if (project == rootProject) {
+        kotlin {
+          ktfmt().googleStyle()
+          licenseHeaderFile(rootProject.file("gradle/license-header-java.txt"), "$")
+          target("*.gradle.kts", "buildSrc/*.gradle.kts", "buildSrc/src/**/kotlin/**")
```
XN137, line 279: i am probably missing something but what is `buildSrc` referring to?
should this be `build-tools` ? it seems to be the only folder containing a `kotlin` subfolder in the repo

XN137 (approved):
+1 since it seems to format the build files correctly (please do it in a separate commit next time to ease reviews...) but not sure if the other kotlin files are also formatted atm (see comment)

PR #226, opened 2022-07-13
Replace incorrect use of getIndex() in EnumValueDescriptor with getNumber()

### LINKED ISSUE #225: Protobuf enums inconsistently use `ordinal` and `getNumber`
In #29 there's an open item to implement more featureful enum handling. However, the [proposal 190](https://github.com/google/cel-spec/wiki/proposal-190) hasn't been approved yet

In the meanwhile, `cel-java` appears to inconsistently use the [ordinal/index of the enum value](https://github.com/projectnessie/cel-java/blob/8e02beb72f08a408b040163f7e86af8b4e9baa9d/core/src/main/java/org/projectnessie/cel/common/types/pb/EnumValueDescription.java#L48) and the [getNumber of the protobuf enum value descriptor](https://github.com/projectnessie/cel-java/blob/75d90ceea7c11d0d9a28ee05804e61da46440e0d/core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java#L267).

This leads to bugs where the enum values are out of order in the protobuf enum declaration.

rbtying:
Fixes #225 

1. [Best practice for protocol buffers](https://errorprone.info/bugpattern/ProtocolBufferOrdinal) is to use `getNumber` rather than `ordinal` on the generated Java enum (which the `EnumValueDescriptor`'s `getIndex` corresponds to
2. The `cel-go` implementation that `cel-java` is based on uses `getNumber`
3. other parts of `cel-java` use `getNumber`

Note: the test update corrects the test to do the right thing, but it doesn't fail in the existing (incorrect) implementation because the test enum has `getNumber() == getIndex()` for all values.

### DIFF
--- a/core/src/test/java/org/projectnessie/cel/common/types/ProviderTest.java
+++ b/core/src/test/java/org/projectnessie/cel/common/types/ProviderTest.java
@@ -69,7 +69,7 @@ import org.projectnessie.cel.common.types.ref.TypeRegistry;
 import org.projectnessie.cel.common.types.ref.Val;
 import org.projectnessie.cel.common.types.traits.Indexer;
 import org.projectnessie.cel.common.types.traits.Lister;
-import org.projectnessie.test.proto3.OutOfOrderEnumOuterClass;
+import org.projectnessie.cel.test.proto3.OutOfOrderEnumOuterClass;
 
 public class ProviderTest {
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/projectnessie/cel/common/types/ProviderTest.java
```diff
@@ -87,12 +88,29 @@ void typeRegistryCopy() {
...
+
+    // Previously, we checked `getIndex` on the `EnumValueDescriptor`, which is the same as the
+    // `ordinal` value on the enum.
+    assertThat(OutOfOrderEnumOuterClass.OutOfOrderEnum.TWO.ordinal())
+        .isEqualTo(OutOfOrderEnumOuterClass.OutOfOrderEnum.TWO.getValueDescriptor().getIndex());
+    // Test the case where the protobuf-defined value for the enum differs from the generated Java
+    // enum's ordinal() function.
+    assertThat(OutOfOrderEnumOuterClass.OutOfOrderEnum.TWO.getNumber())
+        .isNotEqualTo(OutOfOrderEnumOuterClass.OutOfOrderEnum.TWO.ordinal());
+    assertThat(OutOfOrderEnumOuterClass.OutOfOrderEnum.TWO.getNumber())
```
snazy, line 107: Nit: you can add the 2nd `.isEqualTo()` directly behind the one in line 106 (aka: remove the semicolon in 106 and remove line 107).

snazy (commented):
Thanks! The test looks good - maybe just add anohter value.

snazy (approved):
Changes LGTM!
Thanks for the fix!

PR #251, opened 2022-09-02
Add function activation

### LINKED ISSUE #250: Accept variable resolution function instead of map from Script.execute
Hello, first of all thanks for building this project! I'm considering using it for adding dynamic expression support to planetiler yaml config files (https://github.com/onthegomap/planetiler/issues/179). For this I need to construct hierarchical contexts where each nested one adds some global variables beyond what the parent context has.

It looks like the `Activation` utility describes what I'm trying to do, but the `Script#execute` function only takes a `Map`.  Would it make sense for another variation on that to take an `Activation` or even just a `Function<String, Object>` ? I'm happy to make a PR to add this, just want to check if this is the right approach first? Or if there's another way of doing this that I'm missing?

msbarry:
Fix #250 by adding a new variant to `Script.execute` that takes a `Function<String, Object>` instead of `Map<String, Object>`.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
+++ b/core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
@@ -18,6 +18,7 @@ package org.projectnessie.cel.interpreter;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import org.projectnessie.cel.common.types.ref.Val;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
```diff
@@ -114,6 +118,32 @@ public String toString() {
...
+    /** Parent implements the Activation interface method. */
+    @Override
+    public Activation parent() {
+      return null;
+    }
+
+    /** ResolveName implements the Activation interface method. */
+    @Override
+    public Object resolveName(String name) {
+      return provider.apply(name);
```
msbarry, line 138: `MapActivation` has some logic for memoizing lazy values provided by a `Supplier`. `FunctionActivation` lazily computes them by default, but should there be similar logic for memoizing some or all calls to the provider?

snazy (commented):
Looks really neat¹

snazy (approved):

snazy:
Thanks for the contribution!

PR #299, opened 2022-12-15
Allow publishing Gradle build scans from CI

snazy:
But not automatically for every build to protect sensitve workflows, like releasw workflows that have access to sensitive secrets. Gradle build scans are available for free (Gradle Enterprise is not required). The scans help identifying build issues, dependency issues, build performance issues. Also interesting: it shows the result of each test of a CI run.

Build scans are not published for every run - especially sensitive workflows (release workflows w/ access to secrets) are not scanned/published.

Links to the build scans are available from the Summary page of the workflow runs.


dimas-b (approved):
:+1:

PR #368, opened 2023-04-13
Handle `null`/not-present attributes at the "root" level

### LINKED ISSUE #360: Handling nulls in the root fields of the JSON object
Hey team!

I need ability to check `nulls` in the root fields of the JSON object.

Example:
```java
package evaluator;

import com.fasterxml.jackson.core.type.TypeReference;
import com.fasterxml.jackson.databind.JsonNode;
import com.fasterxml.jackson.databind.ObjectMapper;
import org.junit.Test;
import org.projectnessie.cel.checker.Decls;
import org.projectnessie.cel.tools.Script;
import org.projectnessie.cel.tools.ScriptHost;

import java.util.Map;

import static junit.framework.TestCase.assertTrue;

public class JavaCelEvaluator {

  @Test
  public void testRootForNull() throws Exception {
    var celExpression = "foo == null";
    var payloadJson = "{\"foo\":null}";

    var mapper = new ObjectMapper();
    JsonNode jsonNode = mapper.readTree(payloadJson);
    var payloadMap = mapper.convertValue(jsonNode, new TypeReference<Map<String, Object>>() { });

    var scriptHost = ScriptHost.newBuilder().build();
    Script script = scriptHost.buildScript(celExpression)
        .withDeclarations(Decls.newVar("foo", Decls.Any))
        .build();

    var result = script.execute(Boolean.class, payloadMap);

    assertTrue(result);
  }
}
```

This UnitTest is throwing exception:

```
undeclared reference to 'id: 1, attributes: [id: 1, names: [foo]]' (in container '')
org.projectnessie.cel.tools.ScriptExecutionException: 
	at org.projectnessie.cel.tools.Script.evaluate(Script.java:59)
	at org.projectnessie.cel.tools.Script.execute(Script.java:45)
	at com.auvik.platform.alerting.condition.evaluator.JavaCelEvaluator.testRootForNull(JavaCelEvaluator.java:31)
```

The problem is appearing only for root fields, i.e. the same test with this values works great:

```java
var celExpression = "foo.bar == null";
var payloadJson = "{\"foo\":{\"bar\":null}}";
```

The cause of this exception is in the following `null` checks [here](https://github.com/projectnessie/cel-java/blob/fcecfe222d23396bbf8c2362764ed94847f22a15/core/src/main/java/org/projectnessie/cel/interpreter/AttributeFactory.java#L614) and [here](https://github.com/projectnessie/cel-java/blob/fcecfe222d23396bbf8c2362764ed94847f22a15/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java#L1699).

Is it possible to delete those checks? 

If not, how can I check root fields for `null` without changing CEL expression?

P.S. the same unit test with [cel-go](https://github.com/google/cel-go) works fine.

stashkevi4:
### What are you trying to accomplish?

- Fix [the issue with null handling of the root objects](https://github.com/projectnessie/cel-java/issues/360)

### What approach did you choose and why?

The bug was in the way how non-existing attributes were handled:

`object` with value `null` in the root was marked as non-existing, that's why exceptions were raised.

Cel-go returns 2 parameters with `object` (which could be `nil`) and `found`:

`obj, found := vars.ResolveName(nm)` [link to original code](https://github.com/google/cel-go/blob/f564fe1a4e072bca9587913b44a97bd3d38209ca/interpreter/attributes.go#L285)

I've used existing `Pair` generic class as `Pair<Object, Boolean>` to imitate the same behaviour as in Go. 

Now `resolveName` for **MapAttribute** implementation checks that map contains the key, even if obj is null it will be returned as an existing one:

```java
public Pair<Object, Boolean> resolveName(String name) {
      if (!bindings.containsKey(name)) {
        return new Pair<>(null, false);
      }
      Object obj = bindings.get(name);

      if (obj instanceof Supplier) {
        obj = ((Supplier) obj).get();
        bindings.put(name, obj);
      }
      return new Pair<>(obj, true);
    }
```

Another approach was throwing `noSuchAttributeException`, but due to many usages and required `try catches` wrappers code became less readable, so this approach was declined.

Unit tests that covers desired behaviour were added.

These changes are not breaking existing Interpreter UnitTests.

Fixes #360


snazy:
Can you sign the CLA?
And revert the 2nd commit - that's all handled during releases.

core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
```diff
@@ -32,7 +33,7 @@ public interface Activation {
    * ResolveName returns a value from the activation by qualified name, or false if the name could
    * not be found.
    */
-  Object resolveName(String name);
+  Pair<Object, Boolean> resolveName(String name);
```
salvalcantara, line 36: This is very go-idiomatic but does not feel quite right here. In Java, I'd rather use:

```
Optional<Object> resolveName(String name);
```

which reads much better IMO. Indeed, for consistency I would rename that function to:

- `tryResolveName`

core/src/main/java/org/projectnessie/cel/interpreter/AttributeFactory.java
```diff
@@ -353,34 +354,35 @@ public Object resolve(org.projectnessie.cel.interpreter.Activation vars) {
...
         if (typ != null) {
           if (qualifiers.isEmpty()) {
-            return typ;
+            return new Pair<>(typ, true);
           }
           throw noSuchAttributeException(this);
         }
       }
-      return null;
+      return new Pair<>(null, false);
```
salvalcantara, line 385: Instead of returning a value, why not simply `throw noSuchAttributeException(this)` here vs waiting until `resolve`?  This way, `tryResolve` either throws or returns a (potentially null) value. In the latter case you no longer need to distinguish between absence and null.

salvalcantara, line 385: An an alternative, why not fail here (`throw noSuchAttributeException(this)`)? This way, `tryResolve` either succeeds (returning a potentially null value) or fails (throwing an exception). Then `resolve` would need the following adjustments:

```java
    @Override
    public Object resolve(org.projectnessie.cel.interpreter.Activation vars) {
        for (NamespacedAttribute attr : attrs) {
          try {
            return attr.tryResolve(vars);
          } catch (Exception e) {}
        }
       throw noSuchAttributeException(this);
   }
```

stashkevi4: Yeah, there are actually 2 interfaces and I agree that `tryResolve` could be avoided with 1 try catch and exception ignoring.

Another interface is `resolveName`, which I'd rather leave with `Pair<Object, Boolean>` implementation, 'cause it will require put `findIdent` logic inside `catch` section, which could be confusing🤷.

Here's example:
```java
@Override
    public Object tryResolve(org.projectnessie.cel.interpreter.Activation vars) {
      for (String nm : namespaceNames) {
        // If the variable is found, process it. Otherwise, wait until the checks to
        // determine whether the type is unknown before returning.
       try {
           Object op = vars.resolveName(nm);
           for (Qualifier qual : qualifiers) {
             Object op2 = qual.qualify(vars, op);
             if (op2 instanceof Err) {
               return op2;
             }
             if (op2 == null) {
               break;
             }
             op = op2;
           }
           return op;
         }
       } catch (Exception ignore) {
           // Attempt to resolve the qualified type name if the name is not a variable identifier.
           Val typ = provider.findIdent(nm);
           if (typ != null) {
             if (qualifiers.isEmpty()) {
               return typ;
             }
             throw noSuchAttributeException(this);
          }
         }
       }
      throw noSuchAttributeException(this);
    }
``` 

I'm gonna change `tryResolve` with `try catch` approach, it looks really cleaner. 

If  `resolveName` with exception throwing is more preferable, I can change that too :)

core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
```diff
@@ -21,6 +21,7 @@
 import java.util.function.Function;
 import java.util.function.Supplier;
 import org.projectnessie.cel.common.types.ref.Val;
+import org.projectnessie.cel.shaded.org.antlr.v4.runtime.misc.Pair;
```
snazy, line 24: A pair in generally is fine. Just not such a fan of making the relocated antlr api class part of this API.

WDYT about:
```java
package org.projectnessie.cel.interpreter;

import java.util.Objects;

public interface ResolvedValue {
  Object value();
  boolean present();

  ResolvedValue NULL_VALUE = new NullValue();
  ResolvedValue ABSENT = new AbsentValue();
  static ResolvedValue resolvedValue(Object value) { return new PresentValue(value); }

  final class PresentValue implements ResolvedValue {
    private final Object value;
    private PresentValue(Object value) { this.value = Objects.requireNonNull(value); }
    @Override public Object value() { return value; }
    @Override public boolean present() { return true; }
  }

  final class NullValue implements ResolvedValue {
    private NullValue() {}
    @Override public Object value() { return null; }
    @Override public boolean present() { return true; }
  }

  final class AbsentValue implements ResolvedValue {
    private AbsentValue() {}
    @Override public Object value() { return null; }
    @Override public boolean present() { return false; }
  }}
```

The above can lead to more megamorphic call sites, but OTOH there are already a bunch of these in the code base. Alternatively something like this probably serves better?

```java
public final class ResolvedValue {
  public static final ResolvedValue NULL_VALUE = new ResolvedValue(null, true);
  public static final ResolvedValue ABSENT = new ResolvedValue(null, false);
  public static ResolvedValue resolvedValue(Object value) {
    return new ResolvedValue(Objects.requireNonNull(value), true);
  }

  private final Object value;
  private final boolean present;

  private ResolvedValue(Object value, boolean present) {
    this.value = value;
    this.present = present;
  }

  public Object value() { return value; }
  public boolean present() { return present; }
}
```

core/src/main/java/org/projectnessie/cel/interpreter/AttributeFactory.java
```diff
@@ -609,10 +600,9 @@ public Object qualify(org.projectnessie.cel.interpreter.Activation vars, Object
     @Override
     public Object resolve(org.projectnessie.cel.interpreter.Activation vars) {
       for (NamespacedAttribute attr : attrs) {
-        Object obj = attr.tryResolve(vars);
-        // If the object was found, return it.
-        if (obj != null) {
-          return obj;
+        try {
+          return attr.tryResolve(vars);
+        } catch (Exception ignore) {
```
snazy, line 605:
```suggestion
        } catch (ErrException ignore) {
```

snazy (commented):
Thanks a lot!
The changes mostly look good to me. Just a bit concerned about the use of the shaded antlr runtime class and making it part of the CEL-Java "internal-ish" API

core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
stashkevi4: Your solutions are great, I was just trying to use the existing Class and found that Pair :)

I like your second solution more, to keep things simple. But first one works for me too. WDYT?

snazy: Yea, the 2nd one is good enough :tm: ;)

core/src/test/java/org/projectnessie/cel/interpreter/ActivationTest.java
```diff
@@ -50,8 +49,8 @@ void activation() {
   @Test
   void resolve() {
     Activation activation = newActivation(Collections.singletonMap("a", True));
-    assertThat(activation.resolveName("a").a).isSameAs(True);
-    assertThat(activation.resolveName("a").b).isSameAs(true);
+    assertThat(activation.resolveName("a").present()).isSameAs(true);
```
snazy, line 52: Nit: use `.isTrue()`, because the result's a primitive `boolean`
(no need to change the PR though)

core/src/main/java/org/projectnessie/cel/interpreter/AttributeFactory.java
```diff
@@ -17,14 +17,7 @@
...
 import static org.projectnessie.cel.common.types.BoolT.True;
-import static org.projectnessie.cel.common.types.Err.indexOutOfBoundsException;
-import static org.projectnessie.cel.common.types.Err.isError;
-import static org.projectnessie.cel.common.types.Err.maybeNoSuchOverloadErr;
-import static org.projectnessie.cel.common.types.Err.noSuchAttributeException;
-import static org.projectnessie.cel.common.types.Err.noSuchKey;
-import static org.projectnessie.cel.common.types.Err.noSuchKeyException;
-import static org.projectnessie.cel.common.types.Err.noSuchOverload;
-import static org.projectnessie.cel.common.types.Err.throwErrorAsIllegalStateException;
+import static org.projectnessie.cel.common.types.Err.*;
```
snazy, line 20: Please don't use wildcard imports

snazy:
Changes LGTM. If you guys are okay with it, i can go ahead and merge it and get a new release out.

salvalcantara:
I'm fine with the changes too @snazy ! I was just wondering if we could not use `Optional<Object>` instead of the newly defined type: https://github.com/projectnessie/cel-java/pull/368#discussion_r1166789454.

core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
snazy: Not sure (and too sad there's no "null-aware" `Optional`).

A `NULL_VALUE` could mislead other users, because they'd have to do a `if (something != NULL_VALUE)` - so easy to run into issues if they'd do `something != null`, which is different.

snazy (approved):

salvalcantara (commented):
LGTM.

msbarry:
So with this change, how should a `FunctionActivation` provider indicate that it doesn't know how to resolve a name?  Should we change `FunctionActivation(Function<String, Object> provider)` to `FunctionActivation(Function<String, ResolvedValue> provider)` or have some other magic return value from the provider to indicate "absent" or "null"?

stashkevi4:
> So with this change, how should a `FunctionActivation` provider indicate that it doesn't know how to resolve a name? Should we change `FunctionActivation(Function<String, Object> provider)` to `FunctionActivation(Function<String, ResolvedValue> provider)` or have some other magic return value from the provider to indicate "absent" or "null"?

That's a good question. I see 3 different scenarios how it could be handled:
1. Right now, if `provider.apply` result is null - it will raise an Exception (the same as it was before logic change)
2. As you've proposed, we can allow users to set desired value explicitly as `ResolvedValue` instead of `Object` and change provider signature to `Function<String, ResolvedValue>`, but it could cause unforeseen consequences for existing users, that were already sending ``Function<String, Object>`
3. WDYT about implicit implementation of `FunctionActivation` with considering `null` values? Like this one:
```java
/** ResolveName implements the Activation interface method. */
    @Override
    public ResolvedValue resolveName(String name) {
      Object activatedValue = provider.apply(name);
      if (activatedValue == null) {
        return ResolvedValue.NULL_VALUE;
      } else {
        return ResolvedValue.resolvedValue(activatedValue);
      }
    }
```

msbarry:
That seems reasonable, but would a function provider need to differentiate between null and absent? Currently my code does this by returning `null` to indicate absent, and `NullT.NullValue` to indicate it was actually a null value.

stashkevi4:
> That seems reasonable, but would a function provider need to differentiate between null and absent? Currently my code does this by returning `null` to indicate absent, and `NullT.NullValue` to indicate it was actually a null value.

Yeah, in this case only approach with provider signature change to `Function<String, ResolvedValue>` will work, otherwise `null` will be indicating both `absent` and `null`.

msbarry:
Changing the function signature doesn't seem like _too_ big of a concern since this project is pre-1.0.  What do you think @snazy ?

2 possible alternatives:

```java
    @Override
    public ResolvedValue resolveName(String name) {
      Object activatedValue = provider.apply(name);
      if (activatedValue instanceof ResovedValue resolvedValue) {
        return resolvedValue;
      } else {
        return ResolvedValue.resolvedValue(activatedValue);
      }
    }
```

Or we could make another `ResolvedFunctionActivation` or something that takes a `Function<String, ResolvedValue` ?

msbarry:
I put out a proposed fix in #382

PR #382, opened 2023-04-26
Allow function activations to return "absent"

msbarry:
#368 introduced a regression that made it impossible for function activations to indicate that they don't have a value for a name. This change retains the old behavior by default of treating `null` as "absent", but also lets functions return a resolved value directly.

### DIFF
--- a/core/src/test/java/org/projectnessie/cel/interpreter/ActivationTest.java
+++ b/core/src/test/java/org/projectnessie/cel/interpreter/ActivationTest.java
@@ -29,7 +29,10 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
+
+import com.google.protobuf.NullValue;
 import org.junit.jupiter.api.Test;
+import org.projectnessie.cel.common.types.NullT;
 import org.projectnessie.cel.common.types.pb.DefaultTypeAdapter;
 import org.projectnessie.cel.common.types.ref.Val;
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/projectnessie/cel/interpreter/ActivationTest.java
snazy, line 1: Can you add a few changes like these?

In `ResolvedValue` add:
```
  @Override
  public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;

    ResolvedValue that = (ResolvedValue) o;

    if (present != that.present) return false;
    return Objects.equals(value, that.value);
  }

  @Override
  public int hashCode() {
    int result = value != null ? value.hashCode() : 0;
    result = 31 * result + (present ? 1 : 0);
    return result;
  }

  @Override
  public String toString() {
    return "ResolvedValue{present=" + present + ", value=" + value + '}';
  }
```

Replace the added test method by just changing the precedeing test like this?
```
  @Test
  void hierarchicalActivation() {
    // compose a parent with more properties than the child
    Map<String, Object> parentMap = new HashMap<>();
    parentMap.put("a", stringOf("world"));
    parentMap.put("b", intOf(-42));
    parentMap.put("d", stringOf("child value for d"));
    Activation parent = new Activation.FunctionActivation(parentMap::get);
    // compose the child such that it shadows the parent
    Map<String, Object> childMap = new HashMap<>();
    childMap.put("a", True);
    childMap.put("c", stringOf("universe"));
    childMap.put("d", ResolvedValue.NULL_VALUE);
    Activation child = new Activation.FunctionActivation(childMap::get);
    Activation combined = newHierarchicalActivation(parent, child);

    // Resolve the shadowed child value.
    assertThat(combined.resolveName("a")).isEqualTo(ResolvedValue.resolvedValue(True));
    // Resolve the parent only value.
    assertThat(combined.resolveName("b")).isEqualTo(ResolvedValue.resolvedValue(intOf(-42)));
    // Resolve the child only value.
    assertThat(combined.resolveName("c"))
        .isEqualTo(ResolvedValue.resolvedValue(stringOf("universe")));
    // Resolve the child value as null without looking to parent.
    assertThat(combined.resolveName("d")).isSameAs(ResolvedValue.NULL_VALUE);
    // Absent
    assertThat(combined.resolveName("e")).isSameAs(ResolvedValue.ABSENT);
  }
```

And remove the imports (convenience: with `./gradlew spotlessApply`).

snazy (commented):
Nice catch and thanks for the PR! The change looks generally good to me.

Yes, `FunctionActivation` behaved indeed differently after the change introducing `ResolvedValue`.

core/src/test/java/org/projectnessie/cel/interpreter/ActivationTest.java
msbarry: Thanks, sounds good - just added those!

snazy (approved):

snazy:
Thanks for the PR!
I'll go ahead and publish a new release soon.

PR #417, opened 2023-06-30
Add ULong list support for fixed32-like repeated fields

buildbreaker:
What I observed was that a fixed32 gets interpreted as an `int` rather than a `long` during evaluation. 

There are various parts of `FileDescription` that swapped between considering fixed32s as an integer or a long. The change here is to also make things more consistent within each other.


snazy (commented):
Generally LGTM - but need to take a closer look.

buildbreaker:
@snazy Thank you!

core/src/test/java/org/projectnessie/ListContainsTest.java
```diff
@@ -0,0 +1,156 @@
...
+    Map<String, Object> arguments = new HashMap<>();
+    arguments.put("this", 15);
+    arguments.put("rules", rule);
+
+    String result = script.execute(String.class, arguments);
+    assertThat(result).isEqualTo("value must be in list");
+  }
+
+  @Test
+  public void uintInList() throws ScriptException {
```
buildbreaker, line 112: Positive test

core/src/test/java/org/projectnessie/ListContainsTest.java
```diff
@@ -0,0 +1,156 @@
...
+            .build();
+    Map<String, Object> arguments = new HashMap<>();
+    arguments.put("rules", rule);
+
+    Boolean result = script.execute(Boolean.class, arguments);
+    assertThat(result).isTrue();
+  }
+
+  @Test
+  public void uintNotInList() throws ScriptException {
```
buildbreaker, line 135: Negative test

buildbreaker:
@snazy Sorry, the tests are failing. I'll fix them before I ping you for another review.

buildbreaker:
Ok. I believe this is good from my fork's [branch](https://github.com/buildbreaker/cel-java/pull/7)

Let me know if you would like me to move any files around @snazy . Thank you for your help here!

core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java
```diff
@@ -137,14 +138,14 @@ private static Class<?> reflectTypeOfField(FieldDescriptor fieldDesc) {
...
       case SINT32:
-      case FIXED32:
         return Integer.class;
       case INT64:
       case SFIXED64:
       case SINT64:
         return Long.class;
       case UINT32:
       case UINT64:
+      case FIXED32:
```
buildbreaker, line 148: This change is to make reading fixed32 consistent. The other parts of the file considers fixed32s as a ULong (e.g. [here](https://github.com/projectnessie/cel-java/blob/main/core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java#L396))

buildbreaker:
Hey @snazy this is ready for you to take another look when you're available :)

Happy to chat if you have any questions!

core/src/main/java/org/projectnessie/cel/interpreter/AttributeFactory.java
```diff
@@ -1215,6 +1217,12 @@ static Val refResolve(TypeAdapter adapter, Val idx, Object obj) {
       }
       return elem;
     }
+    if (celVal instanceof Lister) {
```
snazy, line 1220: Isn't this `if` superfluous, because `Lister extends Indexer`?

snazy (commented):
Thanks for the change - it looks reasonable and good!
Just one comment about an `if` that seems superfluous.

core/src/main/java/org/projectnessie/cel/interpreter/AttributeFactory.java
buildbreaker: You are right. Let me remove this redundant check

buildbreaker:
@snazy Thanks for the review. I removed the redundant check that you called out!

snazy (approved):

PR #419, opened 2023-06-30
Fix indeterministic hashing for BytesT

buildbreaker:
I'm looking into different list look up use cases and found an issue with bytes. I looked into it a bit and found that what is happening is that the `BytesT` is calling the `super.hasCode()` which is unintentionally calling a `hashCode()` directly on an array which led to an indeterministic hashed value (it's preferred to use `Arrays.hashCode(b)` instead).

The change here is to change the initial value from calling `super.hashCode()` to just using the value 0. The alternative is to use the value `Arrays.hashCode(b)` if preferred.


snazy (approved):
LGTM
Thanks!

PR #426, opened 2023-07-09
Enhanced String Length Handling

elliotmjackson:
This PR introduces improvements to the way string length is handled. The existing `String.length()` method has limitations, and this proposal suggests alternative models for more accurate results. By considering Unicode codepoints and multilingual text, developers can achieve greater flexibility and precision.

### DIFF
--- a/core/src/test/java/org/projectnessie/cel/common/types/StringTest.java
+++ b/core/src/test/java/org/projectnessie/cel/common/types/StringTest.java
@@ -217,4 +217,10 @@ public class StringTest {
     assertThat(stringOf("hello world").size()).isEqualTo(intOf(11));
     assertThat(stringOf("\u65e5\u672c\u8a9e").size()).isEqualTo(intOf(3));
   }
+
+  @Test
+  void stringSizeEmoji() {
+    assertThat(stringOf("").size()).isSameAs(IntZero);
+    assertThat(stringOf("\uD83D\uDE05\uD83D\uDE04\uD83D\uDC7E").size()).isEqualTo(intOf(3));
+  }
 }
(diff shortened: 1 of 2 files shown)


snazy:
Generally no concerns about adding such a functionality. However, this change breaks existing behavior. It would be fine to add it as a _new_ function.

elliotmjackson:
Thanks for your quick reply! I totally see your point about not breaking current behavior. I just wanted to mention that the Java and Go CEL implementations handle string lengths differently, causing potential confusion.

Happy to add this as a new function. Could you give me some pointers on that?

snazy:
So Go's string-length works differently? Interesting.
I found [this issue](https://github.com/google/cel-spec/issues/33), which updates the CEL spec to talk about string-length vs. code-points, so I suspect your current change makes CEL-Java's string-length conformant to the CEL spec. I think, that classifies this PR as a legit bug fix, which then justifies the change in behavior.

Could you add some test-cases to validate the behavior?

elliotmjackson:
sure can, thanks for your thoughtful consideration.

core/src/test/java/org/projectnessie/cel/common/types/StringTest.java
```diff
@@ -217,4 +217,10 @@ void stringSize() {
     assertThat(stringOf("hello world").size()).isEqualTo(intOf(11));
     assertThat(stringOf("\u65e5\u672c\u8a9e").size()).isEqualTo(intOf(3));
   }
+
+  @Test
+  void stringSizeEmoji() {
+    assertThat(stringOf("").size()).isSameAs(IntZero);
+    assertThat(stringOf("\uD83D\uDE05\uD83D\uDE04\uD83D\uDC7E").size()).isEqualTo(intOf(3));
```
elliotmjackson, line 224: is equal to "😅😄👾"

snazy (approved):

PR #476, opened 2023-11-17
Bump cel-spec + googleapi submodules

snazy:
Bump cel-spec submodule to [v0.13.0](https://github.com/google/cel-spec/releases/tag/v0.13.0), googleapi submodule to [f2d78630d2c1d5e20041dfff963e093de9298e4d](https://github.com/googleapis/googleapis/commit/f2d78630d2c1d5e20041dfff963e093de9298e4d).

This was necessary due to recent build errors caused by missing (old) artifacts for the conformance bazel build.

The submodule bumps bring a bunch of new conformance tests as well, which unveiled that some functionality that was not present in CEL-Go when CEL-Java was created, is required by the CEl-Spec.

Summary of changes:
* Bugfix in `o.p.c.checker.Standard` that fixes a type resolution error for equals and not-equals functions, when both parameters need to be up-casted.
* Changes in many types (in `o.p.c.common.types.*T` classes) to automatically convert between numeric and null types.
  * equal and compare with a null and a non-null type no longer fail, but return `False`
  * equal and compare between different numeric types no longer fail, but return "the right" result
  * this includes that numeric CEL map keys can be heterogenous, e.g. an `int` can be retrieved using an `uint` or `double` key
* Fix retrieval of milliseconds from `Duration` - must only return the milliseconds within the second
* Un-ignore a bunch of conformance tests that pass fine now

A few conformance tests have been added to `InterpreterTest`, but the majority of tests is left in the conformance tests.


core/src/main/java/org/projectnessie/cel/common/types/IntT.java
```diff
@@ -216,6 +200,62 @@ public Val convertToType(Type typeValue) {
...
+      case Uint:
+        if (i < 0L) {
+          // this int is negative, so it MUST be smaller than any uint
+          return IntNegOne;
+        }
+        if (other.intValue() < 0L) {
+          // the OTHER uint is > Integer.MAX_VALUE, so THIS int MUST be smaller
+          return IntNegOne;
+        }
+        return intOfCompare(Long.compareUnsigned(i, other.intValue()));
```
dimas-b, line 219: nit: At this point the values are within the `[0; Integer.MAX_VALUE]` range... Do we really need to use `Long.compareUnsigned()`?

dimas-b (approved):

PR #479, opened 2023-11-20
Introduce `cel-standalone` (no dependencies) artifact

snazy:
Generated Protobuf code must be used with the exact same Protobuf runtime version, because especially provides no back-/forward compatibility guarantees. This change introduces a new artifact `cel-standalone`, which provides a shadow/uber jar with including relocated classes of the required dependencies. `cel-standalone` should only be used when there are library compatibility issues, but must not be used in combination with other CEL artifacts.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -48,9 +48,6 @@ idea-ext = { module = "gradle.plugin.org.jetbrains.gradle.plugin.idea-ext:gradle
 immutables-value-annotations = { module = "org.immutables:value-annotations", version.ref = "immutables" }
 immutables-value-processor = { module = "org.immutables:value-processor", version.ref = "immutables" }
 jackson-bom = { module = "com.fasterxml.jackson:jackson-bom", version = "2.16.0" }
-jackson-core = { module = "com.fasterxml.jackson.core:jackson-databind" }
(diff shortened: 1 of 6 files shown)


gradle/libs.versions.toml
```diff
@@ -48,9 +48,6 @@ idea-ext = { module = "gradle.plugin.org.jetbrains.gradle.plugin.idea-ext:gradle
 immutables-value-annotations = { module = "org.immutables:value-annotations", version.ref = "immutables" }
 immutables-value-processor = { module = "org.immutables:value-processor", version.ref = "immutables" }
 jackson-bom = { module = "com.fasterxml.jackson:jackson-bom", version = "2.16.0" }
-jackson-core = { module = "com.fasterxml.jackson.core:jackson-databind" }
-jackson-dataformat-protobuf = { module = "com.fasterxml.jackson.dataformat:jackson-dataformat-protobuf" }
-jackson-dataformat-yaml = { module = "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml" }
```
dimas-b, line 53: Why are we removing this from toml?

dimas-b (approved):
LGTM. Just one question, mostly for my own education :)

gradle/libs.versions.toml
snazy: Just to remove unnecessary decls - don't need manage versions.

PR #519, opened 2024-02-01
CI: Bazel setup action change

snazy:
Force bazelisk to use bazel 6.5.0, because latest 7.0.2 does not work

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,12 +94,6 @@ jobs:
           key: bazel-${{ hashFiles('**/.gitmodules') }}
           restore-keys: bazel-
 
-      - name: Setup bazel
-        if: ${{ matrix.java-version == '11' }}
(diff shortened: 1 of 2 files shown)


dimas-b (approved):

.github/workflows/main.yml
```diff
@@ -94,12 +94,6 @@ jobs:
           key: bazel-${{ hashFiles('**/.gitmodules') }}
           restore-keys: bazel-
 
-      - name: Setup bazel
```
snazy, line 97: Bazelisk is now included in the github-runner image.
Just need to make sure that we're using a Bazel version that works with the conformance tests.

snazy: Note: looks like bazel introduced a _behavioral_ change in v7, that lets the conformance tests build fail.

dimas-b (approved):

PR #584, opened 2024-08-05
Necessary refactor for protobuf v4

snazy:
* Bump submodules cel-spec + googleapis
* Move symlinks in `cel-generated-pb` around
* Use `testFixtures` instead of "test-jar" (less boilerplate)
* Remove jandex plugin (conflicts with the protobuf plugin all the time)


core/src/testFixtures/java/org/projectnessie/cel/Util.java
```diff
@@ -15,13 +15,11 @@
  */
 package org.projectnessie.cel;
 
-import static org.assertj.core.api.Assertions.assertThat;
```
adutra, line 18: Why this change?

snazy: Hm - i only moved the file actually

adutra (approved):

Issue #863, opened 2026-06-05
Partial evaluation behaves incorrectly with  _==_ / _!=_ when one operand is unknown.


clmatt:
During partial evaluation, the equality operators discard unknown operands and return a concrete result.  Specifically an expression like false || x == 'foo' with x unknown returns concrete false when using partial eval, claiming the expression is false for every possible value of x, which is not true. All other operatorsI tested (>, >=, in, &&/|| absorption  comprehensions, function dispatch) propagate unknowns correctly.
  
Reproducing (I tested with 0.6.0): 
```
import java.util.Map;
  import org.projectnessie.cel.Env;
  import org.projectnessie.cel.EnvOption;
  import org.projectnessie.cel.EvalOption;
  import org.projectnessie.cel.Program;
  import org.projectnessie.cel.ProgramOption;
  import org.projectnessie.cel.checker.Decls;
  import org.projectnessie.cel.common.types.ref.Val;
  import org.projectnessie.cel.interpreter.Activation;
  import org.projectnessie.cel.interpreter.AttributePattern;

  public class Repro {
    static Env env;

    static Val partialEval(String expr) {
      Env.AstIssuesTuple parsed = env.parse(expr);
      if (parsed.hasIssues()) throw new IllegalStateException(parsed.getIssues().toString());
      Env.AstIssuesTuple checked = env.check(parsed.getAst());
      if (checked.hasIssues()) throw new IllegalStateException(checked.getIssues().toString());
      Program prg =
          env.program(checked.getAst(), ProgramOption.evalOptions(EvalOption.OptPartialEval));
      Activation.PartialActivation vars =
          Activation.newPartialActivation(
              Map.of(),
              AttributePattern.newAttributePattern("x"),
              AttributePattern.newAttributePattern("y"));
      return prg.eval(vars).getVal();
    }

    public static void main(String[] args) {
      env =
          Env.newEnv(
              EnvOption.declarations(
                  Decls.newVar("x", Decls.String), Decls.newVar("y", Decls.Int)));
      System.out.println("x == 'foo'           -> " + partialEval("x == 'foo'"));
      System.out.println("x != 'foo'           -> " + partialEval("x != 'foo'"));
      System.out.println("y > 5                -> " + partialEval("y > 5"));
      System.out.println("y >= 5               -> " + partialEval("y >= 5"));
      System.out.println("x in ['foo']         -> " + partialEval("x in ['foo']"));
[...]

PR #865, opened 2026-06-09
fix(interpreter): propagate unknown/error operands through == and !=

clmatt:
Using partial evaluate with unknown variables with `==` or `!=` would incorrectly output `true` or `false` as opposed to `unknown`. 

See the issue at https://github.com/projectnessie/cel-java/issues/863. 

For the corrected test, email is unknown, so email == "wiley@acme.co" (and therefore the whole expression) should evaluate to unknown, not false.

### DIFF
--- a/core/src/test/java/org/projectnessie/cel/interpreter/InterpreterTest.java
+++ b/core/src/test/java/org/projectnessie/cel/interpreter/InterpreterTest.java
@@ -1505,6 +1505,32 @@ class InterpreterTest {
     assertThat(result).isInstanceOf(Err.class);
   }
 
+  @Test
+  void equalityWithUnknownOperandStaysUnknown() {
+    assertThat(evalWithUnknownStringX("x == 'foo'")).isInstanceOf(UnknownT.class);
+    assertThat(evalWithUnknownStringX("x != 'foo'")).isInstanceOf(UnknownT.class);
+    assertThat(evalWithUnknownStringX("false || x == 'foo'")).isInstanceOf(UnknownT.class);
+    assertThat(evalWithUnknownStringX("false && x == 'foo'")).isSameAs(False);
+  }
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/projectnessie/cel/interpreter/InterpreterTest.java
```diff
@@ -1505,6 +1505,31 @@ void missingIdentInSelect() {
     assertThat(result).isInstanceOf(Err.class);
   }
 
+  @Test
+  void equalityWithUnknownOperandStaysUnknown() {
+    assertThat(evalWithUnknownStringX("x == 'foo'")).isInstanceOf(UnknownT.class);
+    assertThat(evalWithUnknownStringX("x != 'foo'")).isInstanceOf(UnknownT.class);
+    assertThat(evalWithUnknownStringX("false || x == 'foo'")).isInstanceOf(UnknownT.class);
```
dimas-b, line 1512: would it make sense to add a test case for `false && x == 'foo'` -> `false`?.. or is it covered elsewhere?

dimas-b (dismissed):
Thanks for the contribution, @clmatt ! Changes LGTM 👍 

Let's see if @snazy has any comments.

core/src/test/java/org/projectnessie/cel/interpreter/InterpreterTest.java
clmatt: Good point! One sec

clmatt: 👍

dimas-b (approved):

snazy (approved):
LGTM! Thanks for the contribution @clmatt !

PR #895, opened 2026-07-15
More code examples

snazy:
Adds more examples to the README plus corresponding tests for:
* custom functions
* authz style expressions

Also reorganize the README to improve consumability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,8 +141,11 @@ public class MyClass {
 
 ### Jackson objects
 
-Plain Java and Jackson objects can be used as arguments with the
+Plain Java objects can also be exposed through Jackson's bean/property model by using the
 `org.projectnessie.cel.types.jackson3.Jackson3Registry` from `org.projectnessie.cel:cel-jackson3`.
+Use this registry when the object is not a protobuf message and CEL-Java should read properties the
+same way Jackson would serialize them, including JavaBean getters, records, fields, and Jackson
+annotations such as `@JsonProperty`.
 
 ```java
 import org.projectnessie.cel.types.jackson3.Jackson3Registry;
@@ -253,7 +256,7 @@ evaluation failures should not grant access.
 ### Custom functions
 
 Custom functions can be added by implementing the
-[`org.projectnessie.cel.Library`](https://github.com/projectnessie/cel-java/blob/main/core/src/main/java/org/projectnessie/cel/Library.java)
+[`org.projectnessie.cel.Library`](./core/src/main/java/org/projectnessie/cel/Library.java)
 interface. The interface provides declarations via `List<EnvOption> getCompileOptions()` and
 runtime implementations via `List<ProgramOption> getProgramOptions()`.
 
--- a/README.md
+++ b/README.md
@@ -4,12 +4,35 @@
 [![Maven Central](https://img.shields.io/maven-central/v/org.projectnessie.cel/cel-core)](https://search.maven.org/artifact/org.projectnessie.cel/cel-core)
 
 This is a Java port of the [Common-Expression-Language (CEL)](https://opensource.google/projects/cel).
-
 The CEL specification can be found [here](https://github.com/google/cel-spec).
+
+## Contents
+
+- [Getting started](#getting-started)
+- [Usage](#usage)
+  - [Basic scripts](#basic-scripts)
+  - [Protobuf objects](#protobuf-objects)
+  - [Jackson objects](#jackson-objects)
+  - [Authorization-style expressions](#authorization-style-expressions)
+  - [Custom functions](#custom-functions)
+- [Artifacts](#artifacts)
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/projectnessie/cel/CELTest.java
```diff
@@ -241,6 +242,47 @@ void CustomEnv() {
...
+                        return newErr("invalid arguments to startsWith");
+                      }
+                      return boolOf(
+                          value
+                              .convertToNative(String.class)
+                              .startsWith(prefix.convertToNative(String.class)));
+                    })));
+
+    assertThat(prg.eval(mapOf("resource.name", "projects/example")).getVal()).isSameAs(True);
+    assertThat(prg.eval(mapOf("resource.name", "buckets/example")).getVal()).isSameAs(BoolT.False);
```
XN137: nit: import `False` for consistency with `True` above

tools/src/test/java/org/projectnessie/cel/tools/ScriptHostTest.java
```diff
@@ -163,4 +174,284 @@ void readmeExample() throws Exception {
...
+    assertThat(grants(script, Collections.emptyMap())).isFalse();
+  }
+
+  @Test
+  void authorizationExpressionRuntimeErrorIsNonGranting() throws Exception {
+    Script script = authorizationScript();
+    Map<String, Object> arguments = authorizationArguments("storage.googleapis.com");
+    arguments.remove("resource.name");
+
+    assertThat(grants(script, arguments)).isFalse();
```
XN137, line 252: nit: judging from the test name, this produces a runtime error... but instead of asserting the exception  (like the test below), we are only asserting that our test helper `grants` method returns `false`, which it does for other cases as well).

tools/src/test/java/org/projectnessie/cel/tools/ScriptHostTest.java
```diff
@@ -163,4 +174,284 @@ void readmeExample() throws Exception {
...
+  @Test
+  void customGlobalFunctionErrorIsNonGranting() throws Exception {
+    Script script =
+        ScriptHost.newBuilder()
+            .build()
+            .buildScript("getAttribute(\"error\", \"unknown\") == \"prod\"")
+            .withLibraries(new AttributeLibrary("prod"))
+            .build();
+
+    assertThat(grants(script, Collections.emptyMap())).isFalse();
```
XN137, line 301: this seems to be another case where we collapse exceptions to false, but the behavior could be tested more accurately by asserting the exception.
I understand that in a real "grants" implementation you would want to collapse to "false" in those cases, but not sure about testing.

README.md
```diff
@@ -107,93 +130,52 @@ public class MyClass {
...
     Script script = scriptHost.buildScript("inp.name == checkName")
-        // Variable declarations - we need `inp` +  `checkName` in this example
         .withDeclarations(
-            // types for Jackson need the fully qualified class name 
+            // types for Jackson need the fully qualified class name
             Decls.newVar("inp", Decls.newObjectType(MyInput.class.getName())),
             Decls.newVar("checkName", Decls.String))
-        // Register our Jackson object input type (as a java.lang.Class)
+        // Register the Jackson object input type as a java.lang.Class.
         .withTypes(MyInput.class)
```
XN137, line 165: pardon my ignorance, but for me its not immediate clear what "Jackson objects" means i.e. how is `MyInput` different from a pojo?
the example doesnt make it very clear why `.registry(Jackson3Registry.newRegistry())` is needed imo.

README.md
```diff
@@ -202,51 +184,149 @@ dependencies {
...
+is intentional at the authorization boundary because native conversion failures and unexpected
+evaluation failures should not grant access.
 
-Support for Jackson 2 is similar to Jackson 3, with a few differences:
+### Custom functions
 
-* Use `JacksonRegistry` from `org.projectnessie.cel.types.jackson.JacksonRegistry`
-* Use `org.projectnessie.cel:cel-jackson` dependency instead of `org.projectnessie.cel:cel-jackson3`
+Custom functions can be added by implementing the
+[`org.projectnessie.cel.Library`](https://github.com/projectnessie/cel-java/blob/main/core/src/main/java/org/projectnessie/cel/Library.java)
```
XN137: in other spots we are using relative paths

XN137 (dismissed):
looks like a net-positive with minor comments (feel free to ignore)

tools/src/test/java/org/projectnessie/cel/tools/ScriptHostTest.java
XN137: thanks for the update.
we seem to have kept the `isFalse` assertions, which now basically only tests that our `grants` implementation, collapses exceptions to `false`

XN137 (approved):

PR #900, opened 2026-07-16
Preserve unsigned protobuf map entries

snazy:
Reflective protobuf map extraction rebuilt raw map entries without applying the unsigned normalization used for scalar and repeated fields.

Normalize unsigned numeric map keys and values from the map-entry field descriptors, with coverage for unsigned key and value combinations.

### DIFF
--- a/core/src/test/java/org/projectnessie/cel/common/types/pb/FieldDescriptionTest.java
+++ b/core/src/test/java/org/projectnessie/cel/common/types/pb/FieldDescriptionTest.java
@@ -33,6 +33,7 @@ import com.google.protobuf.Timestamp;
 import com.google.protobuf.Value;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.Collections;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/projectnessie/cel/common/types/pb/FieldDescriptionTest.java
```diff
@@ -145,6 +146,28 @@ void getFrom(GetFromTestCase tc) {
     assertThat(got).isEqualTo(tc.want);
   }
 
+  @Test
+  void getFromUnsignedMapEntries() {
```
XN137, line 151: do we have a similar test for `DynamicMessage` ?

XN137 (dismissed):

core/src/test/java/org/projectnessie/cel/common/types/pb/FieldDescriptionTest.java
snazy: added

XN137 (approved):

PR #902, opened 2026-07-16
Use declaring enum class for Jackson names

snazy:
Jackson enum identifiers used the runtime constant class, which is an anonymous subclass for enum constants with class bodies.

Build identifiers from the declaring enum class in both Jackson adapters and cover lookup/native conversion for a constant-specific class body.


XN137 (approved):
lgtm

we could enforce https://errorprone.info/bugpattern/GetClassOnEnum maybe

snazy:
Hm, looks like there's no errorprone applied at all in CEL :shrug: (just the dependency + plugin declared)

PR #911, opened 2026-07-17
Add allocation-focused CEL benchmarks

snazy:
Add JMH coverage for native value adaptation, Java collection/map inputs, dynamic map literals, and protobuf map/repeated-uint access. These cases expose allocation costs that were not covered by the existing scalar-oriented benchmarks.

### DIFF
--- a/core/src/jmh/java/org/projectnessie/cel/common/types/AdapterAllocationBench.java
+++ b/core/src/jmh/java/org/projectnessie/cel/common/types/AdapterAllocationBench.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright (C) 2021 The Authors of CEL-Java
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.cel.common.types;
+
+import static org.projectnessie.cel.common.types.IntT.intOf;
+import static org.projectnessie.cel.common.types.StringT.stringOf;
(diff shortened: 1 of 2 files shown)


core/src/jmh/java/org/projectnessie/cel/common/types/AdapterAllocationBench.java
```diff
@@ -0,0 +1,187 @@
...
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+import org.projectnessie.cel.common.types.ref.TypeRegistry;
+import org.projectnessie.cel.common.types.ref.Val;
+
+@Warmup(iterations = 1, time = 1500, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 5, time = 300, timeUnit = TimeUnit.MILLISECONDS)
+@Fork(1)
+@Threads(2)
```
XN137, line 50: are we intentionally using > 1 threads when benchmarking allocations?

core/src/jmh/java/org/projectnessie/cel/interpreter/InterpreterAllocationBench.java
```diff
@@ -0,0 +1,248 @@
...
+      switch (kind) {
+        case "mapLookup":
+          program =
+              protoProgram(
+                  "msg.map_string_uint64[key] == target",
+                  Decls.newVar("key", Decls.String),
+                  Decls.newVar("target", Decls.Uint));
+          vars =
+              mapOf(
+                  "msg", protoMessage(size), "key", "key-" + (size - 1), "target", (long) size - 1);
```
XN137, line 165: is it a concern that we pass a long for the `Uint` parameter?
iiuc it means we are not only benchmarking map lookups but also some type conversion

XN137 (approved):

core/src/jmh/java/org/projectnessie/cel/common/types/AdapterAllocationBench.java
snazy: yep

core/src/jmh/java/org/projectnessie/cel/interpreter/InterpreterAllocationBench.java
snazy: not that much of a problem (or perf issue) given that there's also CEL-value conversion/mapping involved.

PR #912, opened 2026-07-17
Cache container candidate name resolution

snazy:
Avoid rebuilding candidate-name strings for repeated namespace resolution on the same container. Cache computed candidates per container while returning a defensive array copy to preserve caller isolation.


XN137 (approved):
i assume all these optimizations are driven by local benchmarks you are running?

snazy:
> i assume all these optimizations are driven by local benchmarks you are running?

yep - IIRC I still have one set of JMH benchmarks sitting on the "big commit stack"

PR #913, opened 2026-07-17
Cache field metadata during checking and planning

snazy:
Avoid repeated provider field metadata lookups within a checker or planner instance. Reuse checked select field metadata when constructing field qualifiers instead of immediately querying the provider again.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
+++ b/core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
@@ -126,6 +126,7 @@ public interface InterpretablePlanner {
     private final Container container;
     private final Map<Long, Reference> refMap;
     private final Map<Long, Type> typeMap;
+    private final Map<String, FieldType> fieldTypes = new HashMap<>();
     private final InterpretableDecorator[] decorators;
 
     Planner(
@@ -249,7 +250,7 @@ public interface InterpretablePlanner {
       FieldType fieldType = null;
       Type opType = typeMap.get(sel.getOperand().getId());
       if (opType != null && !opType.getMessageType().isEmpty()) {
-        FieldType ft = provider.findFieldType(opType.getMessageType(), sel.getField());
+        FieldType ft = findFieldType(opType.getMessageType(), sel.getField());
         if (ft != null && ft.isSet != null && ft.getFrom != null) {
           fieldType = ft;
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
```diff
@@ -273,7 +274,11 @@ Interpretable planSelect(Expr expr) {
         return new EvalTestOnly(expr.getId(), op, stringOf(sel.getField()), fieldType);
       }
       // Build a qualifier.
-      Qualifier qual = attrFactory.newQualifier(opType, expr.getId(), sel.getField());
+      Qualifier qual =
+          fieldType != null
+              ? new AttributeFactory.FieldQualifier(
+                  expr.getId(), sel.getField(), fieldType, adapter)
+              : attrFactory.newQualifier(opType, expr.getId(), sel.getField());
```
XN137, line 281: i assume this change avoids a redundant `provider.findFieldType` call in `AttrFactory.newQualifier` ?
is the field lookup that expensive to be worth kind of breaking the factory abstraction?

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
```diff
@@ -249,7 +250,7 @@ Interpretable planSelect(Expr expr) {
       FieldType fieldType = null;
       Type opType = typeMap.get(sel.getOperand().getId());
       if (opType != null && !opType.getMessageType().isEmpty()) {
-        FieldType ft = provider.findFieldType(opType.getMessageType(), sel.getField());
+        FieldType ft = findFieldType(opType.getMessageType(), sel.getField());
```
XN137, line 253: i assume its not worth caching negative lookup results?
(i have no idea how frequently this gets called and how expensive `provider.findFieldType` is)

XN137 (approved):

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
snazy: It's a bit of repeated work, yea.

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
snazy: Not really, no

PR #918, opened 2026-07-17
Derive Jackson enum values from enum classes

snazy:
Jackson2 enum registration depended on the selected Jackson serializer and could fail when annotations selected a custom serializer.

Derive enum values directly from the raw enum class, matching the Jackson3 path, and cover a custom-serialized enum.

### DIFF
--- a/jackson/src/test/java/org/projectnessie/cel/types/jackson/Jackson2TypeDescriptionTest.java
+++ b/jackson/src/test/java/org/projectnessie/cel/types/jackson/Jackson2TypeDescriptionTest.java
@@ -52,6 +52,7 @@ import org.projectnessie.cel.common.types.pb.Checked;
 import org.projectnessie.cel.common.types.ref.Val;
 import org.projectnessie.cel.types.jackson.types.AnEnum;
 import org.projectnessie.cel.types.jackson.types.CollectionsObject;
+import org.projectnessie.cel.types.jackson.types.CustomSerializedEnum;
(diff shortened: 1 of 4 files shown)


jackson/src/test/java/org/projectnessie/cel/types/jackson/Jackson2TypeDescriptionTest.java
```diff
@@ -129,6 +130,17 @@ void enumConstantSpecificClassBodyUsesDeclaringClassName() {
         .isEqualTo(intOf(EnumWithConstantBody.SPECIAL.ordinal()));
   }
 
+  @Test
+  void customSerializedEnumRegistrationUsesEnumConstants() {
+    JacksonRegistry reg = (JacksonRegistry) newRegistry();
+    reg.enumDescription(CustomSerializedEnum.class);
```
XN137, line 136: nit: at surface level the test looks similar to the `EnumWithConstantBody` above, just wondering if there is any conversion roundtrip we could do to make the usage of the custom serializer "more obvious" in the test ?

XN137 (approved):

jackson/src/test/java/org/projectnessie/cel/types/jackson/Jackson2TypeDescriptionTest.java
snazy: :shrug:

PR #921, opened 2026-07-17
Support nullable optionals and generic collections

snazy:
Jackson property type inference only recognized List as a CEL list-like type and treated Optional as an object-like value. Runtime adaptation had the same List-only limitation and returned Optional values as ordinary objects.

Generalize runtime list adaptation to Collection, unwrap Optional values to either their contained CEL value or null, and teach the Jackson 2 and Jackson 3 type descriptors to infer Collection<T> as list<T> and Optional<T> as T. Add regression coverage for Collection, Set, Optional.of, Optional.empty, and nested Optional values.

### DIFF
--- a/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonTypeDescription.java
+++ b/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonTypeDescription.java
@@ -110,7 +110,7 @@ final class JacksonTypeDescription implements TypeDescription {
       com.google.api.expr.v1alpha1.Type keyType =
           findTypeForJacksonType(type.getKeyType(), typeQuery);
       com.google.api.expr.v1alpha1.Type valueType =
-          findTypeForJacksonType(elementType(type), typeQuery);
+          findTypeForJacksonType(type.getContentType(), typeQuery);
       return Decls.newMapType(keyType, valueType);
     } else if (Collection.class.isAssignableFrom(rawClass)) {
       com.google.api.expr.v1alpha1.Type valueType =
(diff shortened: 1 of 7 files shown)


jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonTypeDescription.java
```diff
@@ -124,6 +127,18 @@ com.google.api.expr.v1alpha1.Type findTypeForJacksonType(JavaType type, TypeQuer
     }
   }
 
+  private JavaType elementType(JavaType type) {
+    JavaType elementType = type.getContentType();
+    if (elementType == null && type.containedTypeCount() > 0) {
+      elementType = type.containedType(0);
```
XN137, line 133: just to confirm: even for Maps `containedType(0)` refers to the value type?
or do Maps always have a non-null `getContentType` ?
do we support raw maps without type parameters?

XN137 (dismissed):

jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonTypeDescription.java
snazy: Nice catch! `containedType(0)` would indeed be the key type for `Map<K,V>`.
Jackson map types always have a non-null `getContentType()` representing the value type, including raw maps, for which Jackson uses `Object` for both key and value.
So the fallback is not reached here and raw maps behave as before.
Still, using `elementType()` for maps is misleading, so I’ll switch that line back to `type.getContentType()`.

XN137 (approved):

PR #923, opened 2026-07-20
Improve interpreter lookup paths

snazy:
Cache protobuf field metadata lookups per registry and avoid duplicate allocation under concurrent access. Reduce map activation lookups for non-null bindings and use checked overload IDs for custom function dispatch.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/common/types/pb/ProtoTypeRegistry.java
+++ b/core/src/main/java/org/projectnessie/cel/common/types/pb/ProtoTypeRegistry.java
@@ -78,6 +78,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import org.projectnessie.cel.common.types.TypeT;
 import org.projectnessie.cel.common.types.ref.FieldType;
 import org.projectnessie.cel.common.types.ref.TypeRegistry;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/projectnessie/cel/common/types/pb/ProtoTypeRegistry.java
```diff
@@ -197,6 +200,11 @@ public Val enumValue(String enumName) {
 
   @Override
   public FieldType findFieldType(String messageType, String fieldName) {
+    String cacheKey = messageType + '\n' + fieldName;
+    return fieldTypeCache.computeIfAbsent(cacheKey, key -> loadFieldType(messageType, fieldName));
```
XN137, line 204: just double checking:
why do we here care about concurrent access whereas there we dont:
https://github.com/projectnessie/cel-java/blob/b45fe3f84b515ea3d8f769813aa5e9a6c2a53899/core/src/main/java/org/projectnessie/cel/checker/Checker.java#L660-L671
?

XN137 (approved):

core/src/main/java/org/projectnessie/cel/common/types/pb/ProtoTypeRegistry.java
snazy: `ProtoTypeRegistry` can be shared, while `Checker` is not.

PR #924, opened 2026-07-20
Plan checked ASTs without wrapper conversion

snazy:
Pass checked expression metadata directly into the interpreter planner during program creation.

This avoids constructing a temporary CheckedExpr on the hot compile-to-program path while preserving the existing checked-expression API.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/CEL.java
+++ b/core/src/main/java/org/projectnessie/cel/CEL.java
@@ -153,8 +153,7 @@ public final class CEL {
     }
     // When the AST has been checked it contains metadata that can be used to speed up program
     // execution.
-    CheckedExpr checked = astToCheckedExpr(ast);
-    p.interpretable = p.interpreter.newInterpretable(checked, decs);
+    p.interpretable = p.interpreter.newInterpretable(ast.getExpr(), ast.refMap, ast.typeMap, decs);
 
     return p;
   }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/projectnessie/cel/interpreter/Interpreter.java
```diff
@@ -37,6 +40,24 @@ public interface Interpreter {
    */
   Interpretable newInterpretable(CheckedExpr checked, InterpretableDecorator... decorators);
 
+  /**
+   * NewInterpretable creates an Interpretable from a checked expression and its metadata without
```
XN137: nit: `newInterpretable` (but maybe javadocs shouldnt repeat the method name at all)

core/src/main/java/org/projectnessie/cel/interpreter/Interpreter.java
```diff
@@ -128,6 +149,19 @@ public Interpretable newInterpretable(
       return p.plan(checked.getExpr());
     }
 
+    /** NewIntepretable implements the Interpreter interface method. */
```
XN137: nit: typo `Intepret` and lowercase `n`
(looks like the `Intepretable` typo exists below as well)

core/src/main/java/org/projectnessie/cel/CEL.java
```diff
@@ -153,8 +153,7 @@ private static Program initInterpretable(
     }
     // When the AST has been checked it contains metadata that can be used to speed up program
     // execution.
-    CheckedExpr checked = astToCheckedExpr(ast);
-    p.interpretable = p.interpreter.newInterpretable(checked, decs);
```
XN137, line 157: is there still a valid use case for the old method? should we remove or deprecate it?

XN137 (dismissed):

core/src/main/java/org/projectnessie/cel/CEL.java
snazy: The existing `CheckedExpr` overload appears useful, it has direct callers, and is the abstract compatibility point for external Interpreter implementations (if those exist :shrug:). The new overload is deliberately a default method.

XN137 (approved):

PR #925, opened 2026-07-20
Reduce CEL collection adaptation allocation

snazy:
Avoid eager list, primitive-array, and repeated unsigned protobuf materialization on value adaptation and field access. Return interpreter-built CEL maps directly and fast-path already adapted CEL maps to skip redundant wrapping work. Add a lazy protobuf map field view so indexed lookup no longer builds a full Java map for every field read.

| Area | Before | After | Result |
| --- | ---: | ---: | --- |
| `nativeToValue(ArrayList)` at 1k entries | 4048 B/op, 0.998 us/op | 32 B/op, 0.008 us/op | Fixed via list-backed `ListT`. |
| `nativeToValue(ListValue)` at 1k entries | 4048 B/op, 1.024 us/op | 32 B/op, 0.013 us/op | Fixed via list-backed `ListT`. |
| `nativeToValue(int[])` at 1k entries | 28200 B/op, 5.331 us/op | 32 B/op, 0.008 us/op | Fixed via primitive-array-backed `ListT`. |
| `nativeToValue(long[])` at 1k entries | 28200 B/op, 5.559 us/op | 32 B/op, 0.007 us/op | Fixed via primitive-array-backed `ListT`. |
| `nativeToValue(double[])` at 1k entries | 28248 B/op, 6.268 us/op | 32 B/op, 0.008 us/op | Fixed via primitive-array-backed `ListT`. |
| `nativeToValue(Map<Val, Val>)` at 1k entries | 95304 B/op, 31.331 us/op | 24 B/op, 2.189 us/op | Fixed via already-CEL map fast path. |
| Dynamic map literal | 1144 B/op, 0.331 us/op | 664 B/op, 0.166 us/op | Fixed via direct wrapped-map construction. |
| Java `ArrayList.exists()` early match at 1k entries | 4320 B/op, 0.893 us/op | 304 B/op, 0.129 us/op | Fixed by avoiding eager input copy. |
| Java `int[].exists()` early match at 1k entries | 28472 B/op, 6.359 us/op | 304 B/op, 0.109 us/op | Fixed by avoiding eager primitive boxing. |
| Java map lookup at 1k entries | 104 B/op, 0.048 us/op | 104 B/op, 0.063 us/op | No generic lazy-map rewrite; normal lookup was already allocation-flat. |
| Protobuf map lookup at 1k entries | 79488 B/op, 29.683 us/op | 31320 B/op, 33.275 us/op | Allocation reduced by lazy protobuf map view; CPU roughly neutral/slower in this short run because lookup scans entries. |
| Protobuf repeated unsigned `exists()` early match at 1k entries | 43592 B/op, 11.079 us/op | 640 B/op, 0.238 us/op | Fixed via lazy unsigned list. |
| Protobuf repeated unsigned `exists()` late match at 1k entries | 339218 B/op, 134.378 us/op | 296202 B/op, 108.904 us/op | Improved, but still dominated by per-element wrapping during full scan. |

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java
+++ b/core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java
@@ -15,6 +15,12 @@
  */
 package org.projectnessie.cel.common.types.pb;
 
+import static org.projectnessie.cel.common.types.BoolT.True;
+import static org.projectnessie.cel.common.types.Err.newTypeConversionError;
+import static org.projectnessie.cel.common.types.Err.noMoreElements;
+import static org.projectnessie.cel.common.types.IntT.intOf;
+import static org.projectnessie.cel.common.types.TypeT.TypeType;
+import static org.projectnessie.cel.common.types.Types.boolOf;
 import static org.projectnessie.cel.common.types.pb.PbTypeDescription.reflectTypeOf;
 import static org.projectnessie.cel.common.types.pb.PbTypeDescription.unwrapDynamic;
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java
```diff
@@ -509,18 +530,204 @@ public static Object getValueFromField(FieldDescriptor desc, Message message) {
...
+    }
+
+    @Override
+    public Object value() {
+      return toJavaMap();
+    }
+
+    @Override
+    public Val contains(Val value) {
+      return boolOf(find(value) != null);
```
XN137, line 612: could this be an optimized variant of `find` without having to convert the entry value?

nit: the method parameter should be called `key` to avoid confusion

XN137 (approved):

core/src/main/java/org/projectnessie/cel/common/types/pb/FieldDescription.java
snazy: Issue is that `find()` returns a `Val` - and errors are communicated as `Err` instances.

XN137: but here we are only checking `!= null` ?
or are you saying checking whether the map contains a key should also fail when reading the value behind the key causes an error?

PR #926, opened 2026-07-20
Make Jackson registry caches concurrent

snazy:
Protect first-use Jackson type and enum discovery when registries are shared across evaluations. Use concurrent maps plus synchronized discovery to avoid unsafe cache mutation without relying on recursive computeIfAbsent.

### DIFF
--- a/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
+++ b/jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
@@ -158,20 +158,11 @@ public final class JacksonRegistry implements TypeRegistry {
     if (ed != null) {
       return ed;
     }
-    ed = computeEnumDescription(clazz);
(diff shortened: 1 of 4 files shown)


jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
XN137, line 164: nit: not from your changes but its odd that this populates `enumMap` while `computeEnumDescription` already does so.
(also `compute` prefix to me suggests it doesnt modify any state)

XN137 (dismissed):

jackson/src/main/java/org/projectnessie/cel/types/jackson/JacksonRegistry.java
XN137: side note:
for `computeTypeDescription` and `typeDescription` the population of the caches is also "spread out"

XN137 (approved):

PR #929, opened 2026-07-21
Specialize macro list folds

snazy:
Recognize map and filter macro fold shapes during planning and build result lists with a mutable evaluation-local buffer.

This avoids repeated accumulator list copying while preserving generic fold fallback for non-standard accumulator references.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
+++ b/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
@@ -28,8 +28,10 @@ import static org.projectnessie.cel.interpreter.Coster.Cost.OneOne;
 import static org.projectnessie.cel.interpreter.Coster.Cost.estimateCost;
 import static org.projectnessie.cel.interpreter.Coster.costOf;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -37,6 +39,7 @@ import org.projectnessie.cel.common.operators.Operator;
 import org.projectnessie.cel.common.types.Err;
 import org.projectnessie.cel.common.types.IterableT;
 import org.projectnessie.cel.common.types.IteratorT;
+import org.projectnessie.cel.common.types.ListT;
 import org.projectnessie.cel.common.types.MapT;
 import org.projectnessie.cel.common.types.Overloads;
 import org.projectnessie.cel.common.types.StringT;
@@ -49,6 +52,7 @@ import org.projectnessie.cel.common.types.traits.Container;
 import org.projectnessie.cel.common.types.traits.FieldTester;
 import org.projectnessie.cel.common.types.traits.Negater;
 import org.projectnessie.cel.common.types.traits.Receiver;
+import org.projectnessie.cel.common.types.traits.Sizer;
 import org.projectnessie.cel.common.types.traits.Trait;
 import org.projectnessie.cel.interpreter.Activation.VarActivation;
 import org.projectnessie.cel.interpreter.AttributeFactory.Attribute;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
```diff
@@ -1108,6 +1112,111 @@ public String toString() {
     }
   }
 
+  final class EvalListFold extends AbstractEval implements Coster {
```
XN137, line 1115: do we need to explicitly handle this in `decDisableShortcircuits` ?
https://github.com/projectnessie/cel-java/blob/f0e77f1ba57cd6b17e6e73a7fa497f3743f8c937/core/src/main/java/org/projectnessie/cel/interpreter/InterpretableDecorator.java#L88-L109

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
```diff
@@ -1108,6 +1112,111 @@ public String toString() {
...
+      }
+      Val foldRange = iterRange.eval(emptyActivation());
+      if (!foldRange.type().hasTrait(Trait.IterableType)) {
+        return Cost.Unknown;
+      }
+      long rangeCnt = 0L;
+      IteratorT it = ((IterableT) foldRange).iterator();
+      while (it.hasNext() == True) {
+        it.next();
+        rangeCnt++;
```
XN137, line 1197: should this try to use `Sizer` first or instead (as we do in `listCapacity`) ?

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
```diff
@@ -615,6 +637,128 @@ Interpretable planComprehension(Expr expr) {
...
+              || (!comprehension.getIterVar().equals(name)
+                  && referencesIdent(comprehension.getLoopStep(), name))
+              || (!comprehension.getAccuVar().equals(name)
+                  && referencesIdent(comprehension.getResult(), name));
+        default:
+          return false;
+      }
+    }
+
+    private final class MacroListFold {
```
XN137: static?

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
```diff
@@ -615,6 +637,128 @@ Interpretable planComprehension(Expr expr) {
...
+      }
+      Expr list = add.getArgs(1);
+      if (list.getExprKindCase() != Expr.ExprKindCase.LIST_EXPR
+          || list.getListExpr().getElementsCount() != 1) {
+        return null;
+      }
+      return list.getListExpr().getElements(0);
+    }
+
+    private boolean isCall(Expr expr, String function, int argCount) {
```
XN137: many of these private helpers could be static afaict

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
```diff
@@ -615,6 +637,128 @@ Interpretable planComprehension(Expr expr) {
...
+          }
+          return false;
+        case COMPREHENSION_EXPR:
+          Comprehension comprehension = expr.getComprehensionExpr();
+          return referencesIdent(comprehension.getIterRange(), name)
+              || referencesIdent(comprehension.getAccuInit(), name)
+              || (!comprehension.getIterVar().equals(name)
+                  && referencesIdent(comprehension.getLoopCondition(), name))
+              || (!comprehension.getIterVar().equals(name)
+                  && referencesIdent(comprehension.getLoopStep(), name))
```
XN137, line 746: claude sonnet claims that we are missing optimizations because we dont also check `getAccuVar` here:
```
  ▎ loopCondition and loopStep are guarded by !iterVar.equals(name) but not !accuVar.equals(name), even though both variables are in scope there (unlike result, which correctly excludes only accuVar). Since all standard macros use 
  ▎ __result__ as their accumulator name, any nested map/filter in a transform will contain __result__ in its own loopStep, causing a false positive here and unnecessarily falling back to EvalFold for the outer comprehension. No 
  ▎ correctness risk — the scoping is fine at runtime — but the optimization is skipped whenever macros are nested.
  ▎
  ▎ || (!comprehension.getIterVar().equals(name) && !comprehension.getAccuVar().equals(name)
  ▎     && referencesIdent(comprehension.getLoopCondition(), name))
  ▎ || (!comprehension.getIterVar().equals(name) && !comprehension.getAccuVar().equals(name)
  ▎     && referencesIdent(comprehension.getLoopStep(), name))
```

XN137 (dismissed):

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
snazy: Nice catch! I already have a follow-up for that.

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
snazy: yes

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
snazy: it's right

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
```diff
@@ -1677,6 +1786,65 @@ public String toString() {
...
+      while (it.hasNext() == True) {
+        iterCtx.val = it.next();
+
+        Val include = fold.filter != null ? fold.filter.eval(iterCtx) : True;
+        Val value = fold.transform.eval(iterCtx);
+        if (include == False) {
+          continue;
+        }
+        if (include != True) {
+          result = noSuchOverload(null, Operator.Conditional.id, include);
```
XN137, line 1821: minor sonnet finding:
```
  EvalExhaustiveListFold.eval() handles errors inconsistently depending on their source.

  Filter errors overwrite result unconditionally:
  if (include != True) {
      result = noSuchOverload(null, Operator.Conditional.id, include);  // always overwrites
      continue;
  }

  Transform errors only set result if it's still null:
  if (result == null) {
      if (isUnknownOrError(value)) {
          result = value;  // only captures the first
      }
  }

  The consequence: if a transform error occurs on element 1, then a filter error occurs on element 2, the transform error is silently replaced by the filter error. Conversely, multiple transform errors always return the first one, while
  multiple filter errors return the last one. There's also no test covering the filter-returns-error path in exhaustive mode at all (both existing exhaustive tests use include == False or a transform that errors, not a filter that returns
  a non-boolean).

  The fix would be to guard the filter-error assignment with the same result == null check:
  if (include != True) {
      if (result == null) {
          result = noSuchOverload(null, Operator.Conditional.id, include);
      }
      continue;
  }

  This is a low-severity issue since exhaustive mode is primarily for cost estimation and observability rather than production evaluation paths, and erroring filter conditions are rare. But the inconsistency is real and would be easy to
  fix.
```

XN137 (approved):

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
snazy: Good catch! I'll keep it in mind!

snazy: Handled in a later follow-up PR

PR #931, opened 2026-07-22
Resolve checked identifiers directly

snazy:
Plan undecorated checked top-level variables as direct activation lookups instead of absolute attributes.

Select roots and decorated programs keep the attribute shape so qualifier planning, partial unknowns, and custom decorators retain their existing behavior.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
+++ b/core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
@@ -55,6 +55,7 @@ import org.projectnessie.cel.interpreter.Interpretable.EvalAttr;
 import org.projectnessie.cel.interpreter.Interpretable.EvalBinary;
 import org.projectnessie.cel.interpreter.Interpretable.EvalEq;
 import org.projectnessie.cel.interpreter.Interpretable.EvalFold;
+import org.projectnessie.cel.interpreter.Interpretable.EvalIdent;
 import org.projectnessie.cel.interpreter.Interpretable.EvalList;
 import org.projectnessie.cel.interpreter.Interpretable.EvalListFold;
 import org.projectnessie.cel.interpreter.Interpretable.EvalMap;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
```diff
@@ -261,7 +266,7 @@ Interpretable planSelect(Expr expr) {
 
       Select sel = expr.getSelectExpr();
       // Plan the operand evaluation.
-      Interpretable op = plan(sel.getOperand());
+      Interpretable op = planSelectOperand(sel.getOperand());
```
XN137: i have to admit that some of these internal concepts and dense names fly over my head, but according to claude sonnet there is an issue here which seems legit (afaict):

**`planSelectOperand` runs unconditionally, dropping qualifier observation in decorated programs**

  `planSelectOperand` is only needed when there are no decorators — but it currently runs regardless, changing behavior for `decObserveEval` and any other decorator.

  Trace `a.b` (checked top-level var) under `decObserveEval`:

  **Before this PR:**
  `plan(a)` → `EvalWatchAttr(EvalAttr_a)`. When `planSelect` calls `attr.addQualifier(qual)`, `EvalWatchAttr.addQualifier` wraps the qualifier `b` in an `EvalWatchConstQual`, so the qualifier's expression id is recorded in `EvalState`.

  **After this PR:**
  `planSelectOperand(a)` → raw `EvalAttr_a`. The qualifier `b` is added unwrapped. The outer `decorate()` call then wraps the whole select in `EvalWatchAttr`, but only the select's id is observed — the qualifier id never is.

  Result: for a select whose operand is a checked top-level ident, **qualifier expression ids are no longer recorded in `EvalState`** in decorated programs. This contradicts the description's claim that "decorated programs … retain their
  existing behavior."

  The fix is a single guard — only take the fast path when there is nothing to decorate:

  ```java
  Interpretable op =
      decorators.length == 0 ? planSelectOperand(sel.getOperand()) : plan(sel.getOperand());
  ```

  Would also be good to add a test: plan a.b (checked-var operand) with a state-tracking decorator and assert the recorded EvalState ids match those produced without this PR.

XN137 (dismissed):

core/src/main/java/org/projectnessie/cel/interpreter/InterpretablePlanner.java
snazy: The concern is valid - will push a fix

XN137 (approved):

PR #934, opened 2026-07-23
Specialize receiver varargs fallback

snazy:
Route unresolved varargs calls through a receiver-specific evaluator that avoids building the full argument array before receiver dispatch.

Also pass the complete receiver tail to dynamic receivers and cover unchecked receiver varargs calls with multiple arguments.

### DIFF
--- a/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
+++ b/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
@@ -1388,11 +1388,6 @@ public interface Interpretable {
     }
 
     private Val receiveOrNoSuchOverload(Val arg0, Val... tailArgs) {
-      for (Val argVal : tailArgs) {
-        if (isUnknownOrError(argVal)) {
-          return argVal;
-        }
-      }
       if (arg0.type().hasTrait(Trait.ReceiverType)) {
         return ((Receiver) arg0).receive(function, overload, tailArgs);
       }
--- a/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
+++ b/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
@@ -822,8 +822,7 @@ public interface Interpretable {
       // Otherwise, if the argument is a ReceiverType attempt to invoke the receiver method on the
       // operand (arg0).
       if (arg0.type().hasTrait(Trait.ReceiverType)) {
-        return ((Receiver) arg0)
-            .receive(function, overload, Arrays.copyOfRange(argVals, 1, argVals.length - 1));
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
```diff
@@ -1318,6 +1317,145 @@ public String toString() {
...
+          return argVal;
+        }
+        tailArgs[i - 1] = argVal;
+      }
+      return receiveOrNoSuchOverload(arg0, tailArgs);
+    }
+
+    private Val receiveOrNoSuchOverload(Val arg0, Val... tailArgs) {
+      for (Val argVal : tailArgs) {
+        if (isUnknownOrError(argVal)) {
```
XN137: nit: this check seems unnecessary as `tailArgs` has been checked for errors already

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
```diff
@@ -1318,6 +1317,145 @@ public String toString() {
...
+          return argVal;
+        }
+      }
+      if (arg0.type().hasTrait(Trait.ReceiverType)) {
+        return ((Receiver) arg0).receive(function, overload, tailArgs);
+      }
+      return noSuchOverload(arg0, function, overload, tailArgs);
```
XN137, line 1394: the existing `EvalVarArgs` calls this with all arguments, not just the tail:
https://github.com/snazy/cel-java/blob/f130b4db348f23e019d98b08fe21a41d7dce620e/core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java#L764C1-L764C37
is this intentional or would this produce different errors?

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
snazy: Nice catch! Updated.

core/src/main/java/org/projectnessie/cel/interpreter/Interpretable.java
snazy: Yep, this produces a different error message, and that's intentional.
`arg0` is already passed separately to `noSuchOverload()` as the receiver, so including it again in `args` duplicates the receiver type in the rendered signature.
Passing only `tailArgs` is consistent with `EvalUnary`, `EvalBinary`, and `Receiver.receive()`.
The existing `EvalVarArgs` behavior looks inconsistent here.

XN137: ok i thought this is outside the "arg0 is receiver" branch, but if the inconsistency with `EvalVarArgs` is fine, then ok

XN137 (approved):

PR #941, opened 2026-07-24
Pure-Java conformance tests

snazy:
The original Bazel-driven build in CEL-Java to run conformance-tests was leveraging the `simple_test` convenience binary to run tests. That binary no longer exists for CEL-Spec 0.25.2 and was replaced with a Java client in the CEL-Spec version bump PR.

This change goes further and removes the Bazel build entirely and replaces the client/server split, which was needed for the `simple_test` tool, with JUnit test-factory.


conformance/src/test/java/org/projectnessie/cel/conformance/SimpleConformanceTest.java
```diff
@@ -0,0 +1,705 @@
...
+          "enums/legacy_proto3/assign_standalone_int_too_big,assign_standalone_int_too_neg",
+          "enums/strong_proto2",
+          "enums/strong_proto3",
+          "fields/qualified_identifier_resolution/map_key_float",
+          "wrappers/uint64/to_json_string",
+          "wrappers/field_mask/to_json",
+          "wrappers/timestamp/to_json",
+          "wrappers/empty/to_json");
+
+  private static final Set<String> matchedSkips = new LinkedHashSet<>();
```
XN137, line 182: sonnet finding:

>  AtomicInteger is used correctly for the counters, but matchedSkips is an unsynchronized LinkedHashSet. JUnit 5 @TestFactory tests inside a DynamicContainer can execute concurrently under @Execution(CONCURRENT). Even without that
  annotation, test engines may parallelize by default in some configurations. Concurrent add() calls on a LinkedHashSet produce undefined behavior. Should be:

>  private static final Set<String> matchedSkips = Collections.synchronizedSet(new LinkedHashSet<>());

>  or just a ConcurrentHashMap.newKeySet() (which sacrifices insertion order but that's fine here since it's only used for membership tests).

conformance/src/test/java/org/projectnessie/cel/conformance/SimpleConformanceTest.java
```diff
@@ -0,0 +1,705 @@
...
+  private static final AtomicInteger total = new AtomicInteger();
+  private static final AtomicInteger passed = new AtomicInteger();
+  private static final AtomicInteger skipped = new AtomicInteger();
+
+  @TestFactory
+  Stream<DynamicNode> simpleConformance() {
+    List<DynamicNode> files = new ArrayList<>();
+    TEST_FILES.forEach(fileName -> files.add(dynamicContainer(fileName, fileTests(fileName))));
+    files.add(dynamicTest("skip list matches testdata", this::assertAllSkipsMatched));
```
XN137, line 191: sonnet finding:

>   This sentinel is appended to the file list and relies on JUnit executing it after all dynamic tests have populated matchedSkips. JUnit does not guarantee ordering of dynamic nodes under concurrent execution. If anything parallel is ever
  enabled this will produce false positives. A safer approach would be to collect matchedSkips from the SkipList.parse() result and do the check structurally (compare SKIP_TESTS against the parsed testdata at startup rather than tracking
  hits at runtime).

conformance/src/test/java/org/projectnessie/cel/conformance/SimpleConformanceTest.java
```diff
@@ -0,0 +1,705 @@
...
+    return files.stream();
+  }
+
+  @AfterAll
+  static void printSummary() {
+    System.out.printf(
+        "Conformance tests: %d total, %d passed, %d skipped%n",
```
XN137, line 198: i assume we confirmed we pass and skip the same amount as previously

conformance/README.md
```diff
@@ -1,57 +1,40 @@
...
 ```
 
-After the build succeeds once from that directory, you can use the `./run-conformance-tests.sh`
-script.
+Unmatched skips fail the test suite, so stale skips are visible when upstream
+testdata changes.
+
+Optional CEL-Spec files such as extension libraries, optionals, and type
+deduction are intentionally not enabled by default. Add those separately with
+explicit skip reasoning.
```
XN137, line 40: is the last sentence a TODO ?

XN137 (approved):
+1 assuming we dont want to support concurrent execution

conformance/src/test/java/org/projectnessie/cel/conformance/SimpleConformanceTest.java
snazy: Yes - it'll be eventually be even more

conformance/README.md
snazy: Yes, scope to follow-up work.

PR #943, opened 2026-07-27
CEL spec 0.25.2 conformance

snazy:
This PR makes CEL-Java conformant to the CEL 0.25.2 specification.

Many of the the individual changes, left as separate commits, touch the same files.

The end result is that effectively all CEL 0.25.2 conformance are enabled and pass.

Strong-enum support is intentionally not enabled, because that's not part of the language specification and are mutually incompatible with the enum-as-int conformance sections.


core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
```diff
@@ -236,6 +245,9 @@ public Activation parent() {
 
     @Override
     public ResolvedValue resolveName(String name) {
+      if (name.startsWith(".")) {
+        return delegate.resolveName(name);
```
XN137: are we missing `.substring(1)` here?
otherwise the `if` is useless

core/src/main/java/org/projectnessie/cel/extension/MathLib.java
```diff
@@ -0,0 +1,432 @@
...
+    return Long.compare(left.intValue(), right.intValue());
+  }
+
+  private static double asDouble(Val value) {
+    if (value instanceof DoubleT) {
+      return value.doubleValue();
+    }
+    if (value instanceof UintT) {
+      return Long.toUnsignedString(value.intValue()).equals("18446744073709551615")
+          ? 18446744073709551615.0
```
XN137: why do we need this extra logic?
```
        double a = Double.parseDouble("18446744073709551615");
        double b = 18446744073709551615.0;
        System.out.println(a);
        System.out.println(b);
        System.out.println(a == b);
```
yields:
```
1.8446744073709552E19
1.8446744073709552E19
true
```

core/src/main/java/org/projectnessie/cel/extension/NetworkLib.java
```diff
@@ -0,0 +1,547 @@
...
+  public List<ProgramOption> getProgramOptions() {
+    return List.of(
+        ProgramOption.functions(
+            Overload.unary(IP, NetworkLib::ip),
+            Overload.unary("ip_string", NetworkLib::ip),
+            Overload.unary(CIDR, NetworkLib::cidr),
+            Overload.unary("cidr_string", NetworkLib::cidr),
+            Overload.unary(IS_IP, NetworkLib::isIp),
+            Overload.unary("is_ip_string", NetworkLib::isIp),
+            Overload.unary("is_ip_cidr", value -> newErr("no such overload")),
```
XN137: why is this an error?

core/src/main/java/org/projectnessie/cel/common/types/TypeT.java
```diff
@@ -154,11 +163,11 @@ public boolean equals(Object o) {
     if (this == o) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
+    if (!(o instanceof Type)) {
       return false;
     }
     Type typeValue = (Type) o;
-    return typeEnum == typeValue.typeEnum() && typeName().equals(typeValue.typeName());
+    return typeName().equals(typeValue.typeName());
```
XN137, line 173: is it correct to now only rely on the name instead of also the `typeEnum` ?
can there not be conflicts?

core/src/main/java/org/projectnessie/cel/interpreter/Activation.java
snazy: valid cleanup. But `.substring(1)` would be wrong for hierarchical delegates, because the delegate must see the absolute-name marker.

core/src/main/java/org/projectnessie/cel/extension/MathLib.java
snazy: Don't recall, removed the special case.

core/src/main/java/org/projectnessie/cel/common/types/TypeT.java
snazy: CEL runtime type-value equality intentionally compares qualified type names. `typeEnum` is an implementation category so including it would break well-known object-type equality. I documented that invariant in TypeT.java.

core/src/main/java/org/projectnessie/cel/extension/NetworkLib.java
snazy: Meh, nonsense, removed.

XN137 (approved):

PR #964, opened 2026-08-20
Remove `jmh` from CI and `check`

snazy:
Do not run JMH in CI, it's results are not monitored and JMH results from GH hosted runners aren't reliable either.

Also remove JMH invocation from Gradle's `check`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,17 +43,10 @@ jobs:
         with:
           cache-read-only: ${{ github.event_name != 'push' || github.ref != 'refs/heads/main' }}
 
-      - name: Spotless Check
-          # Spotless must run in a different invocation, because
-          # it has some weird Gradle configuration/variant issue
-        if: ${{ matrix.java-version == '21' }}
-        run: ./gradlew spotlessCheck --scan
-
       - name: Build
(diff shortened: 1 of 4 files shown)


snazy:
JMH takes quite a while now and will, with the upcoming changes, get more benchmarks.
Leaving JMH runs for manual invocation.

.github/workflows/main.yml
```diff
@@ -43,17 +43,10 @@ jobs:
...
-          # Spotless must run in a different invocation, because
-          # it has some weird Gradle configuration/variant issue
-        if: ${{ matrix.java-version == '21' }}
-        run: ./gradlew spotlessCheck --scan
-
       - name: Build
         run: >
           ./gradlew --rerun-tasks ${{ env.ADDITIONAL_GRADLE_OPTS }}
           assemble check publishToMavenLocal
-          -x jmh -x spotlessCheck -x testNative
```
XN137, line 56: why do we remove `-x testNative` ? it still seems to get executed separately in the next step

snazy: `testNative` isn't part of `check`

XN137 (approved):

PR #982, opened 2026-09-17
Code style and Java language level adoptions

snazy:
* Replace `Arrays.asList()` with single element with `List.of()`
* enhanced switch expression
* pattern variable
* Use String.repeat()
* Use `.isEmpty()`
* Remove unused `ParseError`
* Remove unused function in TypeErrors
* Remove duplicate `if`


core/src/main/java/org/projectnessie/cel/common/CELError.java
```diff
@@ -102,9 +102,7 @@ public String toDisplayString(Source source) {
       // sophisticated way, maybe use jline's WCWidth, but that one is also quite rudimentary wrt
       // code-blocks (e.g. doesn't know about emojis).
       result.append("\n | ");
-      for (int i = 0; i < location.column(); i++) {
-        result.append(dot);
-      }
+      result.append(String.valueOf(dot).repeat(Math.max(0, location.column())));
```
XN137, line 105: are these kind of changes really an improvement in terms of readability/performance?

core/src/main/java/org/projectnessie/cel/parser/ParseError.java
```diff
@@ -1,31 +0,0 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package org.projectnessie.cel.parser;
-
-import org.projectnessie.cel.common.Location;
-
-public final class ParseError extends RuntimeException {
```
XN137, line 20: does this break backward-compatibility of our public api?
if yes maybe need to document this in the release notes

XN137 (approved):

core/src/main/java/org/projectnessie/cel/common/CELError.java
snazy: :shrug:

core/src/main/java/org/projectnessie/cel/parser/ParseError.java
snazy: Nah, not breaking. This `ParseError` type wasn't used at all.

